PR #5, opened 2018-12-19
Feature/inner jsx transformer

kotielnikov:
Added typescript transformer plugin which exposes inner jsx structure of the component.

For now doesn't work with the components which are wrapped into HOCs.


FlorianRappl:
I think the tooling should either be all in js or all in ts. With this PR it will be mixed. Can we make this consistent please? Also is there any reason for not using `ttsc` and going with a custom runner? Besides that looks good to me.

Finally, please sign the CLA - thanks! :beers:

src/components/FileUploaderDetails/UploaderProgressBar.part.tsx
```diff
@@ -1,13 +1,15 @@
 import * as React from 'react';
+
+import { Icon, IconName } from '../Icon';
+import { ProgressStatus, TranslationLabels } from './FileUploaderDetails.types.part';
+import { defaultLabels, iconNames } from './helpers';
 import styled, { reStyled } from '../../utils/styled';
-import { distance } from '../../distance';
+
```
FlorianRappl: I am not sure what's up with these empty lines in here (and other files). Can you please remove them? Why change the import blocks?

src/components/Spinner/index.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
+
```
FlorianRappl: Same as above.

webpack.config.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const webpack = require('webpack');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
+const createInnerJsxTransformer = require('typescript-plugin-inner-jsx').default;
```
FlorianRappl: This and other lines would not be necessary if we would just use `ttsc`. But maybe you tried that already.

However, I would prefer if we move this all to a configuration in the tsconfig (which is what `ttsc` allows us to do).

kotielnikov: `ttsc` won't work in this case. Transform needs to be applied **before** `emit` step, while by using `ttsc` it can be applied only **during** different steps of `emit`.

The transform needs to be applied before emit because otherwise updated declarations are not generated.
`afterDeclarations` can't be used because on this step the input data provided to the transformer contains declarations only and not the full source, so the transformer can't build inner structure from it.

src/components/FileUploaderDetails/UploaderProgressBar.part.tsx
kotielnikov: I have plugin in vs code which alphabetically sorts imports.

FlorianRappl: Yep, but can you either turn it off or configure it to not insert blank lines? We always have a single block of imports. Also we have a sorting that usually does in order

- all (asterisk) imports
- default imports
- named imports

where external packages `'foo'` are ranked higher than internal ones (`'./foo'`). I have no problem with changing the order slightly, but inserting these blank lines is rather disturbing because I (personally) think some meaningful LOC starts already.

webpack.config.js
FlorianRappl: Okay understood. But then I think we should not have the build.ts here. Rather, it should be already in the ts transform package, because others will also need this runner then. It should come in form of an executable (i.e., register in bin) such that it may be executed directly (at least from npm scripts / or if installed globally / via npx / ...).

kotielnikov: Don't agree. I believe that transformer package should contain transformer only and it's integration into the system is up to the consumer. We can only advice how to integrate it.
Transformer itself follows all the principles and does it's job perfectly - transforms code from A to B. Compiler definitely doesn't belong to the package because it serves different purpose. And the reason why we need a custom compiler here because:

- typescript doesn't have standard way to run transformers through tsc arguments or tsconfig
- we need to change order transfomer runs
- we need not only js files but d.ts as well

...And the end consumer may not have these reasons

FlorianRappl: That does not make much sense. We just supply the standard runner - we do not force anyone to use it. However, any implementation should look similar to ours, so not providing this default implementation would be a waste. Even if someone would want to roll out his/her own implementation - it would be not different than now. So

- users who are fine with our implementation don't have to do anything
- other users are not affected

There is no reason to not roll it out in the typescript transform package. Again, it does not force anyone to use it.

FlorianRappl: Just to be clear - what the transformer package does is pretty much what we want from the runner. Of course, we want js in the end. Of course, we want to have the d.ts affected (otherwise its only half of a job). Of course, we want it in the stage you describe.

Again, if someone has something else in mind including it in the package already will not have a negative impact here. There are only advantages.

package.json
```diff
@@ -12,8 +12,8 @@
     "test": "npm run lint && npm run test:unit",
     "prettier": "prettier --config prettier.config.js --write 'src/**/*.{ts,tsx}'",
     "build:bundle": "cross-env NODE_ENV=production webpack -p",
-    "build:es5": "tsc -t es5 --outDir dist/es5 --declaration --inlineSourceMap --inlineSources",
-    "build:es6": "tsc -t es6 -m es6 --outDir dist/es6 --declaration --inlineSourceMap --inlineSources",
+    "build:es5": "node_modules/typescript-plugin-inner-jsx/bin/build -t es5 --outDir dist/es5 --declaration --inlineSourceMap --inlineSources --config ./tsconfig.json",
```
FlorianRappl: Hm if you properly register it you can just call something like `tsc-transform-jsx`  (the name is up to you)

FlorianRappl: Please see, e.g., https://medium.com/netscape/a-guide-to-create-a-nodejs-command-line-package-c2166ad0452e (section 3) for how to properly register a script in package.json (this is missing in typescript-plugin-inner-jsx).

PR #7, opened 2019-01-11
ProgressBar Component #36181

boromo:
### Prerequisites

### Contribution Type

### Description

ProgressBar component updated based on the latest design updates. One additional property `type` was added with two possible options: `primary` (default) and `secondary`.


FlorianRappl:
Changed to branch to `devel` @boromo - we need to go to `devel` first!

src/components/ProgressBar/index.tsx
```diff
@@ -83,29 +98,36 @@ const ProgressIndicator = styled.div`
...
   position: absolute;
-  width: 12px;
-  height: 12px;
-  top: -4px;
-  background: ${themed(props => (props.active ? props.theme.fill : props.theme.textDisabled))};
-  border: 2px solid ${themed(props => props.theme.background)};
-  transform: translate(-50%, 0);
+  width: ${distance.small};
+  height: ${distance.small};
+  background: ${({ active }) => (active ? midnight : grey4)};
```
FlorianRappl: Are these colors also available on the theme? Should we take them from the theme (uiX, textX or supportX)?

src/components/ProgressBar/index.tsx
```diff
@@ -124,22 +150,24 @@ function spinning(child: React.ReactChild) {
...
   minimum = 0,
   maximum = 100,
-  value = 0,
+  value: propValue,
   title,
   description,
   animate,
   stops = [],
-  ...props
+  type = 'primary' as 'primary',
```
FlorianRappl: Is this as really necessary? I think the original type of `type` already constraints the string to one of the two values, i.e., `primary` should be of the right type. Maybe I am wrong here.

boromo: Without assigning default value `type`'s type is fine, but with the default value, it's changed to type string. Not sure how can I fix this.

src/components/ProgressBar/index.tsx
boromo: We have `midnight ` used in button's theme, but it doesn't make sense to take it from there. And grey4 is actually not used in the theme.

FlorianRappl: Well, these themes are deprecated and we do not want to use them (best case -> they are all removed). So I am exclusively referring to uiX, textX and supportX. Maybe we should introduce something like that on theme for midnight and grey4, because I can imagine that these will break consistency when custom themes are applied.

boromo: I've added these two colors to theme, but we'll have to review the naming conventions with designers, since I don't like that we use `supportX` color in progress bar, I think the name should be more neutral if they use the same color for support and progress bar.

src/themes.ts
```diff
@@ -150,6 +151,7 @@ export const light: PreciseFullTheme = {
   notificationTextFontSize: remCalc('16px'),
   notificationTextLineHeight: `22px`,
   notificationIconMarginRight: distance.medium,
+  support1: colors.midnight,
```
FlorianRappl: midnight is ui7 -> see zeplin.

src/themes.ts
```diff
@@ -96,6 +96,7 @@ export const light: PreciseFullTheme = {
   ui4: colors.lighterGray,
   ui5: colors.tuna,
   ui6: colors.transparent,
+  ui7: colors.grey4,
```
FlorianRappl: Hm maybe let's do it like that. Use grey directly (grey is grey), but midnight via ui7. I am not sure if grey should be put on the theme as its also missing (like *all* greys) in the pattern library.

src/components/ProgressBar/index.tsx
boromo: Yes, I saw it, but wasn't actually sure which one to use, and that's why I just used support1, but anyways this will be ui7 now, and I'm checking with Emanuel which name to use for this grey one.

PR #20, opened 2019-02-04
Bugfix/missed accordion table labels +  Sustain `Flyout` scroll position on component's update. 

kotielnikov:


src/components/Flyout/FlyoutWindow.part.tsx
```diff
@@ -348,17 +354,38 @@ export class FlyoutWindowInt extends React.Component<FlyoutWindowProps, FlyoutWi
   private setFlyoutRef = (el: HTMLDivElement) => {
     if (el) {
```
FlorianRappl: Potential leak (maybe also apparent somewhere else in the library) - if the callback is used multiple times (either for the same DOM node or - even worse - for multiple nodes) we connect multiple handlers and only disconnect the last one that has been set.

Idea: Always before appending call a function that just executes

```js
    if (this.flyoutContainer) {
      this.flyoutContainer.removeEventListener('scroll', this.onScroll);
    }
```

FlorianRappl:
Looks good to me - I would just respect the versioning we have (i.e., we have not yet released 0.5) and have a look at our use of refs w.r.t. adding / removing event handlers. Potentially we can improve here.

Thanks! Good job! :beers:

src/components/Flyout/FlyoutWindow.part.tsx
kotielnikov: I moved it to componentDidMount

src/components/Flyout/FlyoutWindow.part.tsx
```diff
@@ -366,7 +365,10 @@ export class FlyoutWindowInt extends React.Component<FlyoutWindowProps, FlyoutWi
   };
 
   componentDidMount() {
-    this.updateMeasurements();
+    if (this.flyoutContainer) {
+      this.updateMeasurements();
+      this.flyoutContainer.addEventListener('scroll', this.onScroll);
```
FlorianRappl: Note - this is not a fix for the problem I described. With this solution you face two problems:#

1. If the DOM node changes during the lifetime of the component you no longer receive any updates (and disconnect the event from the wrong DOM node since your this.flyoutContainer is different in the end from the first version)
2. If no DOM node was available at mount (edge case and impl. dependent) then we never connected to the event

FlorianRappl: Yes, but the DOM node can change during runtime.

As a simple (straight forward) example consider (it is one example, there are others where the node change is more hidden):

```jsx
class Hello extends React.Component {
  componentDidMount() {
	  console.log('cdm');
  }
  render() {
    return React.createElement(this.props.el, {
	    ref: el => console.log(el)
    }, `Hello ${this.props.name}`);
  }
}

class Carrier extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      el: 'i',
    };
  }
  componentDidMount() {
    setTimeout(() => {
    	this.setState({ el: 'b' });
    }, 2000);
  }
  render() {
    return <ul><li><Hello name="World" el={this.state.el} /></li></ul>;
  }
}

ReactDOM.render(
  <Carrier />,
  document.getElementById('container')
);
```

What you see here is 1. ref called (initial node) 2. did mount called 3. ref called (null) 4. ref called (new node).

TL;DR: You can go with the implicit mode as currently outlined, but for various robustness reasons I would propose to work against the ref node (if global node (e.g., this.flyoutContainer) is existing disconnect the listener(s), if local node (e.g., el) is existing connect the listener(s)) as it is resilient against all edge cases and not just one happy flow.

src/components/Flyout/FlyoutWindow.part.tsx
```diff
@@ -354,6 +354,8 @@ export class FlyoutWindowInt extends React.Component<FlyoutWindowProps, FlyoutWi
   private setFlyoutRef = (el: HTMLDivElement) => {
     if (el) {
       this.flyoutContainer = el;
+      this.flyoutContainer.removeEventListener('scroll', this.onScroll);
```
FlorianRappl: This wont work (if we really changed the node you are using the new el - which hasnt any listener appended yet).

Hopefully the following code fragment shows you the essential idea I tried to explain:

```jsx
//...

private setRef = (el: Element) => {
  if (this.container) {
    this.container.removeEventListener(...);
  }

  if (el) {
    el.addEventListener(...);
  }

  this.container = el;
}
```

(no cleanup in `componentWillUnmount` needed in this case)

src/components/Flyout/FlyoutWindow.part.tsx
```diff
@@ -352,11 +352,15 @@ export class FlyoutWindowInt extends React.Component<FlyoutWindowProps, FlyoutWi
...
+    }
+
     if (el) {
       this.flyoutContainer = el;
-      this.flyoutContainer.removeEventListener('scroll', this.onScroll);
-      this.flyoutContainer.addEventListener('scroll', this.onScroll);
     }
+
+    this.flyoutContainer.addEventListener('scroll', this.onScroll);
```
FlorianRappl: I'm not sure why the proposed logic was not taken. This one is not symmetric and not as robust as the proposed one.

Can you explain why this version?

kotielnikov: Both setting container and adding listener should occur only if `el` was passed. I've updated the code

FlorianRappl: But if the last time something was passed in `el` was `null` it should be reflected via the global container. Otherwise you

- still have the ref that needs to be considered on the GC (just unnecessary load on the GC as the ref count is still 1, where it could be 0 improving the gen. cleanup)
- reflects a false state of the DOM in the component (even though it should only be `null` when the component unmounts it could still be used)

The logic is just more sound this way. It "somewhat" works already as currently implemented.

PR #24, opened 2019-02-06
Add inner structure to the components wrapped in HOCs

kotielnikov:
### Prerequisites

- [ ] I have read the **CONTRIBUTING** document

### Contribution Type

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "tslint-config-prettier": "^1.17.0",
     "tslint-plugin-prettier": "^2.0.1",
     "typescript": "3.2.2",
-    "typescript-plugin-inner-jsx": "^0.1.6",
+    "typescript-plugin-inner-jsx": "^0.1.7",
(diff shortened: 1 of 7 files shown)


FlorianRappl:
LGTM!

package.json
```diff
@@ -90,7 +90,7 @@
     "tslint-config-prettier": "^1.17.0",
     "tslint-plugin-prettier": "^2.0.1",
     "typescript": "3.2.2",
-    "typescript-plugin-inner-jsx": "^0.1.6",
+    "typescript-plugin-inner-jsx": "^0.1.7",
```
FlorianRappl: Only remark I have - this is a dev dependency and wont be installed. However, we reference the module in our code. Is this reference stripped away in the resulting js? Because if not it will lead to some errors when consuming precise-ui!

FlorianRappl: @kotielnikov please ensure this is not the case (I haven't checked - maybe you already have).

PR #25, opened 2019-02-06
reStyled type fix

kotielnikov:
### Prerequisites

### Contribution Type

### Description

Fix explained in example:

```
interface Props {
    highlighted: boolean;
}
const StyledLink = reStyled<Props, 'a'>('a')((props) => `color: ${props.highlighted ? 'red' : 'blue'}`);

// Before the fix typescript compiler will complain about onClick prop, but onClick is 
// a standard prop of 'a' component

<StyledLink highlighted onClick={() => {}}/>
```

### DIFF
--- a/src/utils/styled.ts
+++ b/src/utils/styled.ts
@@ -41,7 +41,7 @@ export type ThemedStyledFactories<FullTheme, PropsTheme> = {
 export interface ThemedStyled<FullTheme, PropsTheme> extends ThemedStyledFactories<FullTheme, PropsTheme> {
   <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): (
     interpolation: PreciseInterpolationFunc<P, FullTheme>,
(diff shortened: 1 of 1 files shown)


FlorianRappl:
@kotielnikov I have nothing against this PR - it looks good. But I have some weird gut feeling regarding the other thing with the inner-jsx. That we require (as a dependency) a tooling-only lib does not make much sense to me.

My impression was that the import in the HOCs is / should only be there to indicate something to our inner-jsx tool. Then this tool should remove the import and modify the AST accordingly. My weird gut feeling is telling me that we bundle code that was not meant to be bundled.

kotielnikov:
@FlorianRappl I wasn't able to come up with other idea how to point to the plugin that "something" should have inner structure and cover all the cases

FlorianRappl:
Alright @kotielnikov maybe I can come up with an idea. If something bright comes to me I'll ping you and let's discuss.

PR #28, opened 2019-02-07
withInner import path update

kotielnikov:
### Prerequisites

### Contribution Type

### Description

Import paths of `withInner` were updated because of circular dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "dependencies": {
     "memoize-one": "^5.0.0",
     "react-onclickoutside": "^6.7.1",
-    "typescript-plugin-inner-jsx": "^0.1.7"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
   "dependencies": {
     "memoize-one": "^5.0.0",
     "react-onclickoutside": "^6.7.1",
-    "typescript-plugin-inner-jsx": "^0.1.7"
+    "typescript-plugin-inner-jsx": "^0.1.9"
```
FlorianRappl: Even though this is definitely better it does not remove the dependency on the one who installs the lib.

Our ultimate goal should be that this dependency is no longer needed (as we deliver the transpiled sources).

kotielnikov: I will continue thinking about that. Thanks

PR #30, opened 2019-02-12
Feature/modal animations

kotielnikov:
### Prerequisites

### Contribution Type

### Description

Transition animations added and new props.

### DIFF
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -29,16 +29,6 @@ export interface ModalProps extends StandardProps {
    */
   onClose?(e: ModalCloseEvent): void;
   /**
-   * Show or hide close modal button
-   * @default true
-   */
-  showCloseButton?: boolean;
-  /**
-   * Determines if modal should be closed when clicked outside or Esc pressed
-   * @default true
-   */
-  closeOnOutsideClick?: boolean;
-  /**
    * When specified, default max-width of 500px will be overridden
    * @deprecated Please define through styled components and ${Modal.inner.ModalContent}
    */
@@ -218,18 +208,17 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
   };
 
   render() {
-    const { children, onClose, open = false, minHeight, showCloseButton, closeOnOutsideClick, ...rest } = this.props;
-    const showClose = showCloseButton === undefined || showCloseButton;
-    const closeOnOutside = closeOnOutsideClick === undefined || closeOnOutsideClick;
+    const { children, onClose, open = false, minHeight, ...rest } = this.props;
+    const canClose = typeof onClose === 'function';
     const { closing } = this.state;
 
     return (
       open && (
-        <StyledBlocker closing={closing} onClose={closeOnOutside ? this.closeBackground : undefined}>
+        <StyledBlocker closing={closing} onClose={this.closeBackground}>
           <StyledModal tabIndex={0} closing={closing} {...rest}>
             <ModalContent minHeight={minHeight}>
               {children}
-              {showClose && <CloseButton onClick={this.closeButton} />}
+              {canClose && <CloseButton onClick={this.closeButton} />}
             </ModalContent>
           </StyledModal>
         </StyledBlocker>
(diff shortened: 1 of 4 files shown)


src/components/Modal/index.tsx
```diff
@@ -28,12 +28,24 @@ export interface ModalProps extends StandardProps {
    * Event triggered when the modal should be closed by the user.
    */
   onClose?(e: ModalCloseEvent): void;
+  /**
+   * Show or hide close modal button
+   * @default true
+   */
+  showCloseButton?: boolean;
```
FlorianRappl: Not sure if this is needed, e.g., if `onClose` is unset then the button should do nothing --> do not display the button then. Otherwise, I would always tend *not to* introduce *by default* `true` booleans -> they should be `false` such that they are implicit. As such I would (personally) call such a prop (if there really should be one) `hideCloseButton`.

src/components/Modal/index.tsx
```diff
@@ -28,12 +28,24 @@ export interface ModalProps extends StandardProps {
...
+  /**
+   * Determines if modal should be closed when clicked outside or Esc pressed
+   * @default true
+   */
+  closeOnOutsideClick?: boolean;
```
FlorianRappl: Looks fine to me (however, same argument as above) even though I think it is redundant. Keep in mind that the `onClose` allows the user to handle if it should be closed. We have a background origin to indicate that the modal should be closed due to clicking outside. I would not introduce too many redundant properties (especially if they are less powerful / flexible).

kotielnikov: I think if Modal specification has onClose animation, this should be handled by modal itself and not from outside. In this case implementation seems to be pretty complicated but overall it's as simple as possible for this need and pretty logical. onClose handler called after all animations played and modal is ready to be deleted from DOM

kotielnikov:
@FlorianRappl I think if Modal specification has onClose animation, this should be handled by modal itself and not from outside. In this case implementation seems to be pretty complicated but overall it's as simple as possible for this need and pretty logical. onClose handler called after all animations played and modal is ready to be deleted from DOM

src/components/Modal/index.tsx
kotielnikov: removed

src/components/Modal/index.tsx
kotielnikov: removed either

FlorianRappl:
> @FlorianRappl I think if Modal specification has onClose animation, this should be handled by modal itself and not from outside. In this case implementation seems to be pretty complicated but overall it's as simple as possible for this need and pretty logical. onClose handler called after all animations played and modal is ready to be deleted from DOM

I agree that the animation should be in the modal, however, we already have the open prop, so you can be sensitive to this (and being sensitive here in terms of an animation should be fairly straight forward - you just toggle some CSS by having conditional CSS depending on the `open` prop, the rest is done via some transition). I did not propose to move the animation out - just to not have a closing state in the modal as I don't see the need for it (but again you have thought this more through than I have so maybe I am just missing something).

The problem I see (but this is also not touched here) is that we quite often did not use `open` but rather something like `{modalOpen && <Modal open>...}`, which of course will never result in the "closing" animation (the whole thing would be just removed). However, this problem is a cosmetic one and can be touched incrementally (no need for immediate change of everything).

kotielnikov:
when `open` prop is false , modal is removed from DOM and there is no anymore an element to which an animation can be applied. And I don't want to keep modal in dom only because of animation

FlorianRappl:
> when `open` prop is false , modal is removed from DOM and there is no anymore an element to which an animation can be applied. And I don't want to keep modal in dom only because of animation

I see - makes sense!

(Remark: I thought about this as well, e.g., what if we have some modals in the portal they all would be needed to be kept in the DOM; the only solution I would be able to come up here is an API breaking change --> `onClose` becomes `onClosing` where a user could decide *not* to allow closing it - then `onClose` would be only called to notify that we have now closed, e.g., after the animation. Alternative to `onClosing` / `onClose` is `onClose` / `oClosing` or -- maybe the best -- have `onClosing`, `onClose`, `onClosed`, where `onClose` is for-backward compatibility and `onClosing` would trigger the animation / `onClosed` the cleanup; but only in combination with the prop of course. All maybe too complicated / confusing - just wanted to share these thoughts.)

kotielnikov:
so will you approve?

FlorianRappl:
Sure sorry - thought I already did :). :beers:

PR #34, opened 2019-02-19
FileUploaderDetails custom translation

boromo:
## FileUploaderDetails custom translation

### Contribution Type

### Description

Currently we have the problem, that when a consumer application uses the FileUploaderDetails, custom error messages do not get localized, only the known labels, and this feature is allowing consumers to define their custom translate function which would be called with the error string passed to the file object.

### DIFF
--- a/src/components/FileUploaderDetails/Example.md
+++ b/src/components/FileUploaderDetails/Example.md
@@ -14,6 +14,7 @@ class FileUploaderDetailsExample extends React.Component {
     this.cancelUpload = this.cancelUpload.bind(this);
     this.makeProgress = this.makeProgress.bind(this);
     this.onClose = this.onClose.bind(this);
+    this.errorShown = false;
   }
 
   uploadStart({ files: startedFiles }) {
@@ -36,7 +37,7 @@ class FileUploaderDetailsExample extends React.Component {
 
     if (file) {
       const uploaderId = file.uploaderId;
-      const timeout = (Math.floor(Math.random() * Math.floor(3)) + 1) * 1000;
+      const timeout = (Math.floor(Math.random() * 3) + 1) * 1000;
       const currentProgress = file.progress || 0;
 
       if (currentProgress < 100 && !file.canceled && !file.error) {
@@ -46,13 +47,27 @@ class FileUploaderDetailsExample extends React.Component {
         if (file.progress < 100) {
           setTimeout(() => this.makeProgress(id), timeout);
         } else {
-          this.eventManager.emit(FileUploadActions.uploadSuccess, { files: [file] });
+          if (!this.errorShown) {
+            this.errorShown = true;
+            this.eventManager.emit(FileUploadActions.uploadFailure, { files: [{...file, error: 'upload_204'}] });
+          } else {
+            this.eventManager.emit(FileUploadActions.uploadSuccess, { files: [file] });
+          }
         }
       }
 
       this.files[file.fileId] = file;
     }
   };
+
+  translate(error) {
+    switch(error) {
+      case 'upload_204':
+        return 'Localized message for error 204';
+      default:
+        return `No localization for ${error}`;
+    }
+  }
 
   onClose() {
     this.files = {};
@@ -67,6 +82,7 @@ class FileUploaderDetailsExample extends React.Component {
           onUpload={this.uploadStart}
           onCancel={this.cancelUpload}
           onClose={this.onClose}
+          translate={this.translate}
         />
         <FileUploader
           multiple
--- a/src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
+++ b/src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
@@ -73,3 +73,9 @@ export interface FileUploaderDetailsEvent<T> {
    */
   files: Array<T>;
 }
+export interface Translate {
+  /**
+   * Optionally provide custom translate function to be used for unhandled custom error messages
+   */
+  translate?(str: string): string;
+}
--- a/src/components/FileUploaderDetails/UploaderProgressDetails.part.tsx
+++ b/src/components/FileUploaderDetails/UploaderProgressDetails.part.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import styled from '../../utils/styled';
 import { getPropLabel, UploadProgressDetailsLabels } from '../../utils/labels';
 import { Modal, ModalBody, ModalFooter, ModalHeader } from '../Modal';
-import { FileProgress, FileUploaderDetailsEvent } from './FileUploaderDetails.types.part';
+import { FileProgress, FileUploaderDetailsEvent, Translate } from './FileUploaderDetails.types.part';
 import { StatusTable } from './StatusTable.part';
 import { IconLink } from '../IconLink';
 
@@ -15,7 +15,7 @@ const StyledModal = styled(Modal)`
   max-width: 600px;
 `;
 
-export interface UploaderProgressDetailsProps extends UploadProgressDetailsLabels {
+export interface UploaderProgressDetailsProps extends UploadProgressDetailsLabels, Translate {
   /**
    * Determines if the details are shown or not.
    */
--- a/src/components/FileUploaderDetails/StatusTable.part.tsx
+++ b/src/components/FileUploaderDetails/StatusTable.part.tsx
@@ -5,7 +5,7 @@ import { Icon } from '../Icon';
 import { ProgressBar } from '../ProgressBar';
 import { Table, TableCellEvent, TableRowEvent } from '../Table';
 import { ActionIconContainer } from './ActionIconContainer.part';
-import { FileProgress, FileUploaderDetailsEvent } from './FileUploaderDetails.types.part';
+import { FileProgress, FileUploaderDetailsEvent, Translate } from './FileUploaderDetails.types.part';
 import { StatusIcon } from './StatusIcon.part';
 import { getStatus, iconNames } from './helpers';
 import { StandardProps } from '../../common';
@@ -56,7 +56,7 @@ export const StyledTableRow = reStyled.tr<StyledTableRowProps>(
 `,
 );
 
-export interface StatusTableProps extends StandardProps {
+export interface StatusTableProps extends StandardProps, Translate {
   files: Array<FileProgress>;
   onCancel(e: FileUploaderDetailsEvent<FileProgress>): void;
   onDelete(e: FileUploaderDetailsEvent<FileProgress>): void;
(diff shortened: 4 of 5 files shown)


src/components/FileUploaderDetails/Example.md
```diff
@@ -46,14 +47,28 @@ class FileUploaderDetailsExample extends React.Component {
...
       }
 
       this.files[file.fileId] = file;
     }
   };
 
+  translate(error) {
+    switch(error) {
+      case 'upload_204':
+        return 'Localized message for error 204';
```
FlorianRappl: What error is error 204? It seems like this has nothing to do with precise ui. Maybe model error not as a string, but rather as a function of the error? Then in the standard implementation (this one) we could make this switch, otherwise it would be left to the user to write something meaningful based on an error object.

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
```diff
@@ -73,3 +73,9 @@ export interface FileUploaderDetailsEvent<T> {
    */
   files: Array<T>;
 }
+export interface Translate {
+  /**
+   * Optionally provide custom translate function to be used for unhandled custom error messages
+   */
+  translate?(str: string): string;
```
FlorianRappl: Since this is only used for the error I would call it error?(e: Error): string and place it on the general translations object (it should be consistent with the overall approach of Precise UI to localization / messages).

src/components/FileUploaderDetails/UploaderProgressDetails.part.tsx
```diff
@@ -15,7 +15,7 @@ const StyledModal = styled(Modal)`
   max-width: 600px;
 `;
 
-export interface UploaderProgressDetailsProps extends UploadProgressDetailsLabels {
+export interface UploaderProgressDetailsProps extends UploadProgressDetailsLabels, Translate {
```
kotielnikov: Translate can be a part of `StandartProps`

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
kotielnikov: the idea is to introduce a unified way to provide translations to the components which need translations without introducing  new props for each translation (how its currently done in `Accordion`: there are `openLabel` and `closeLabel`: the idea is to have `translate` only which will be able to translate these labels ). not for **errors** rather for any potential text

src/components/FileUploaderDetails/Example.md
kotielnikov: this is just an example, there could be any text

src/components/FileUploaderDetails/UploaderProgressDetails.part.tsx
FlorianRappl: You mean these?

```ts
export interface StandardProps {
  /**
   * Places the given class on the element.
   */
  className?: string;
  /**
   * An optional theme which can be passed down to a component.
   */
  theme?: PreciseTheme;
  /**
   * The style prop for explicitly overriding some CSS styles.
   */
  style?: React.CSSProperties;
}
```

I would not place `translate` here. Most of our components do not use this prop. `StandardProps` is an interface that makes sense for *all* components (e.g., every component consumes `style`, `className`, ... in one way or the other).

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
kotielnikov: I think `str` should be `any`

src/components/FileUploaderDetails/UploaderProgressDetails.part.tsx
kotielnikov: ok, agreed

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
FlorianRappl: The unified way we have in form of labels - which is imho much better as it transports right away to a user what is possible and what is not there.

As it is proposed in this PR I would not know what to translate and what not 😕. It would all be a guessing game until the next component comes and shows me that it may also have something to translate 🔮 .

Right now (with labels) it is very explicit and allows from a central point to set the translation globally (instead of forcing the user to do it per component).

❗️ **Just to be clear** that we have labels at all is already a bit of a mistake 🤒 ; normally there should not be anything that shows a default string. But since Precise UI also comes with some quite bulky (or *useful*, depending on the POV) components such labels have been necessary 🚀 .

kotielnikov: the rule of thumb can be used: you have some text in component, ie "Open details", `translate` will be called with the same text, than:
```
const translate = (str) => {
   if(str === 'Open details') {
        return `Открыть детали`
   }

   return str;
}
<Accordion translate={translate} />
```

Also `Accordion` can export:
```
export type AccordionTranslations = 'Open details' | 'Close details'
```
```
export type Translations = AccordionTranslations | LoaderTranslations | any;
export interface Tranlsate {
  (e: Translations): string
}
```

FlorianRappl: Well, for once I *would* call such a function `getText` or `getLabel` (because `translate` assumes we want to translate, which is only an edge case - usually we only want to override the original wording with our wording).

Second I would not transport "original text", but rather keys (this avoids the duplication problem, which can happen across components and inside components, plus limits the verbosity of this concept).

Finally, I would then not use `string` (or `any`) as the type, but rather either something like `LabelKeys = 'label1' | 'label2' | ...`, or even better some overloads to allow a case with two or more arguments like the described error case, where we have (e.g.,) a key "error" and a custom argument of type `Error`.

Personally, I am still not convinced that this solves a real problem. I think the current approach is much more React-like (we have a prop and the name of the prop is very explicit) plus user-friendly (we can locally override it, or even make the global settings different to already start with a default).

I think there are much severe problems that would need attention than this (still unsure why an inconsistent approach is introduced - either change it completely or go with the existing one).

kotielnikov: > Finally, I would then not use `string` (or `any`) as the type, but rather either something like `LabelKeys = 'label1' | 'label2' | ...`, or even better some overloads to allow a case with two or more arguments like the described error case, where we have (e.g.,) a key "error" and a custom argument of type `Error`.

I didn't get this one. Could you please provide an example especially for the case when component needs to translate something which is not defined in the component (error code).

Consistency can be fixed. Likely we don't have much cases.

src/components/FileUploaderDetails/Example.md
boromo: As Maksym mentioned, it's an example, but it's actually the example based on the issue we have currently. Currently when the error is emitted from the consumer application, the status table just displays the error, and then the consumer has no chance of localizing it. Like in the example above, the service can throw different types of error, and we could use this new function to localize the errors based on e.g. error code returned from the service.

FlorianRappl: > As Maksym mentioned, it's an example, but it's actually the example based on the issue we have currently.

This issue can be also solved by adding an `error` function to the `labels` (passing in the exact error / error object). It also is consistent with the current approach.

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
kotielnikov: Component user wants not to `getText`, `getLabels` but rather set it.

FlorianRappl: > Component user wants not to `getText`, `getLabels` but rather set it.

This is not how APIs are designed imho. It is never about *what the user wants*, because the API does not (and cannot!) know that. The API can only communicated *what it offers* / wants, e.g., the *onClick* is the event *from* the component, but the user *wants* to call an API as a consequence. Still we do not name the event *onCallTheApi*. 

In this case the API offers the user to get the text to display (not saying these are good names, but the get / set discussion is not very useful imho).

(also setting implies that we actually place the value somewhere, when in fact we should only retrieve a value)

src/components/FileUploaderDetails/Example.md
kotielnikov: Approach with `setLabels` have a big problem right now - it doesn't allow to define labels per component. We can introduce `error` function that we will need `notification` and so on,  or we can have a standard function responsible for all translations

I doubt that precise-ui user will be using all components, which means that the scenario where he needs to provide all texts through `setLabel` is much more rare than he need to provide texts to the concrete component. 

The concrete component will have then `setTexts` prop with a type 
```
interface SetTexts {
  (str: 'openLabel' | 'closeLabel'): string
}
```

FlorianRappl: No I think you are wrong. The current approach allows both.

Please check the use right now - you have the globals, which *can* be set using *setLabels* and are used as *fallbacks* when the labels are *not set* on the component.

> I doubt that precise-ui user will be using all components, which means that the scenario where he needs to provide all texts through `setLabel` is much more rare than he need to provide texts to the concrete component.

I know at least one user / use that *uses* all components. And I think you also know this application ... so I wonder why you think there is no such case?

src/components/FileUploaderDetails/FileUploaderDetails.types.part.ts
kotielnikov: For me as consumer `precise-ui` is a black box. I want  to set text of `Card` component to show "Открыть" instead of open. What is more logical prop name for me `setText` or `getText`?

FlorianRappl: For me personally `getText`, because I should not set text, but provide text. Either way what does `translate` do? Should I now translate? I have a feeling we go in circles.

This is not - yet - about any naming. It was just an example. I think we should not have this discussion at all, because the whole PR is not sound.

The problem you are facing can be solved with the current approach easily. So what is the problem?

(this already seems to be about taste instead of logic and the technical basis - I would appreciate if we could tackle the technical foundation first before we make this a discussion on names)

src/components/FileUploaderDetails/Example.md
kotielnikov: `setLabels` for providing globals and fallbacks still can be kept with only extending type by `{[nonDefinedLabel: string]: string}`. this solves one problem

But there is second one - ### texts per component. In fact we already have a problem with this: texts to `FileUploader` currently can be provided globally only through the `setLabels`. There is at least 16 labels related to `FileUploader`. With current approach if I would like to customize `FileUploader` I will need to introduce 16 new props for labels which is just insane.

If we have setLabels globally why just not to have setLabels exclusively  for the component as well?

PR #39, opened 2019-03-04
Styled components upgrade

kotielnikov:
### Prerequisites

### Contribution Type


docs/components/ComponentsList.tsx
```diff
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import { ActionLink, Expander, styled, distance, Icon, ActionLinkProps, themed, reStyled } from '../../src';
+import { ActionLink, Expander, styled, distance, Icon, ActionLinkProps } from '../../src';
 
 // @ts-ignore
 import getUrl from 'react-styleguidist/lib/utils/getUrl';
+import { themed } from '../../src/utils/styled';
```
FlorianRappl: Sure that themed is no no exported? I mean Zeisslets will / may depend on it. Same for reStyled.

docs/components/StyleGuideRenderer.tsx
```diff
@@ -42,33 +42,34 @@ function useTheme() {
...
-          return (
-            <Layout logo={<Logo />} ribbon={<Ribbon />} {...props}>
-              {!window.location.hash ? <HomePage /> : children}
-            </Layout>
-          );
-        }}
-      />
-    </ThemeProvider>
+    <>
+      {window.location.port === '6061' && <GlobalStyle />}
```
FlorianRappl: Woot :) Haven't seen this one. Why do we need this rule? What am I missing?

docs/components/Wrapper.tsx
```diff
@@ -3,7 +3,7 @@ import { ThemeProvider } from '../../src';
 
 const ThemeWrapper: React.SFC = ({ children }) => {
   const { theme } = (window as any).context;
-  return <ThemeProvider theme={theme}>{children}</ThemeProvider>;
+  return <ThemeProvider theme={theme} children={children as any} />;
```
FlorianRappl: Not sure why this needs to be changed / rewritten. Both forms are equivalent and having the children as the children is the preference (only exception would be a prop forwarding where it makes more sense than going over destructuring).

package.json
```diff
@@ -44,7 +44,7 @@
   "peerDependencies": {
     "react": "^16.8.0",
     "react-dom": "^16.8",
-    "styled-components": "^2.x"
+    "styled-components": "^4.1.3"
```
FlorianRappl: I think `^4.x` would be sufficient for the *peer dependency*. This looks more like the *dev dependency*.

src/utils/index.ts
```diff
@@ -13,4 +13,22 @@ export * from './displayTo';
...
+  ThemeProvider,
+  ThemeConsumer,
+  ThemeContext,
+  isStyledComponent,
+  ServerStyleSheet,
+  StyleSheetManager,
+  themed,
+  /**
+   * @deprecated
+   * use `styled` instead
```
FlorianRappl: Hm could also be added to Pharos UI in the compat layer and removed from Precise UI.

docs/components/StyleGuideRenderer.tsx
kotielnikov: do disable animations (for visual testing)

FlorianRappl: Okay thanks good to know (alternatively could have been made work with an environment variable used when building for visual testing or a query parameter with a name like `visual_testing`)!

kotielnikov: may change with the next PR, I have some improvements for visual testing

PR #57, opened 2019-03-22
Typography final

kotielnikov:
### Prerequisites

### Contribution Type

### Description

Typography


src/components/Headline/HeadlineTextStyles.ts
```diff
@@ -0,0 +1,113 @@
+import { remCalc } from '../../utils/remCalc';
+
+export enum HeadlineTextTransform {
```
FlorianRappl: Sure you want to export it as enum - and not const enum? Difference is that const enum will result in a replacement at compile-time, while enum will have a runtime cost.

src/components/Headline/HeadlineTextStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+  none = 'none',
+  upperCase = 'uppercase',
+  lowerCase = 'lowercase',
+  capitalize = 'capitalize',
+  unset = 'unset',
+  initial = 'initial',
+  inherit = 'inherit',
+}
+
+export enum HeadlineTextStyles {
```
FlorianRappl: const enum vs enum again.

src/textStyles.ts
```diff
@@ -1,113 +1,134 @@
...
+export function getFontStyle({ size, weight, lineHeight = size }: FontStyleProps) {
+  return css`
+    ${getFontSize(size)}
+    ${getFontLineHeight(lineHeight)}
+    ${getFontWeight(weight)}
+  `;
+}
+
+/**
+ * @deprecated
```
FlorianRappl: In my opinion these deprecated things can be specified in Pharos - keeping Precise clean.

PR #82, opened 2019-04-26
"withValidation" HOC didn't work at all (fix)

dimabory:
### Prerequisites

### Contribution Type

### Description

I guess an actual issue appeared in https://github.com/ZEISS/precise-ui/pull/78. @kotielnikov Probably, there is a much better solution for this.

#### Before:
v5 0 1

#### After:
v fixed


src/components/Autocomplete/index.tsx
```diff
@@ -282,7 +282,7 @@ class AutocompleteInt<T> extends React.Component<AutocompleteProps<T> & FormCont
       inputRef: _7,
       ...props
     } = this.props;
-    const { open, listFocus, value, error } = this.state;
+    const { open, listFocus, value, error = this.props.error } = this.state;
```
kotielnikov: In that and other similar cases logic should be moved to `componentWillReceiveProps`.
Be careful: currently it's updating state only if component is in controlled mode. Instead it should update `error` on each prop change (not only in controlled mode)

dimabory:
> Can we cover this functionality with some tests to prevent regression here? Thanks!

I apologize for the delayed response because I've just come back from the vacation.
I will add tests as soon as I have more free time for that.

src/components/Autocomplete/index.tsx
dimabory: Check it now.

Issue #126, opened 2019-07-18
Support grouping of items with no value for the 'groupBy' column in AccordionTable


lestard:
### Description

When some of the items in a `AccordionTable` don't have a value for the column that the table is grouped by, the behaviour is not consistent and not very convinient. 

If an item without a value is the first in the data list, then the grouping of items doesn't work correctly at all. In this case items without a value are mixed in between grouped items. In this case it can look like some items without a value are included in one of the groups which is not correct. The source for this problems is in the sorting algorithm that is deciding on how to sort items based on the type of value of the first item in the list.

My proposal includes:
- fix the sorting by not looking at the first item in the list but search for the first item that actually has a value set for the `sortby`/`groupBy` column
- change the order of items so that items without a value are moved to the end of the list
- add a pseudo-group for all items that don't have a value for the `groupBy` field so that you can toggle this group like normal groups.
- Add a prop to `AccordionTable` to define a label value that should be used for this group of items without a value. This prop should be optional. If no value was defined then use something like "-" as a default

guiherzog:
This has been added in the latest commits. Thanks @lestard for the fix.

PR #127, opened 2019-07-19
Feature/126 accordion table sort and group no value fields

lestard:
### Prerequisites

### Contribution Type

### Description

See #126. 

- fix sorting of items that don't have a value for the `sortBy`/`groupBy` column
- change order of items so that items without a value are moved to the end of the list
- `AccordionTable`: 
	- add a separate pseudo-group for items that don't have a value for the `groupBy` column
	- add an optional prop to define the label for this new pseudo-group
- change sorting for values of type number so that `0` isn't handled as `falsy` anymore but instead is sorted and grouped like other numbers

Breaking changes: At the moment the handling of Tables with items that don't have a value defined for the `sortBy`/`groupBy` column isn't working correctly so I assume that most people don't have not-defined values in their data. However, as this PR changes the way not-defined values are handled in theory it could change the behaviour of existing apps that depend on the (wrong) sorting and grouping of not-complete items.


guiherzog:
Hi @lestard, thanks for this PR. There are a few issues with it though:

1. You should use a long dash (—) instead of a short dash (-) on case of an empty group without a name. (In the code, they all look the same, but render differently in the browser).
2. Elements with value "0" should not be included in the empty group because this may cause undesired behaviour.
3. IMO, I believe empty arrays should also be included in the empty group and, as of now, they are being included to their own not name group.
4. The groups also show as the first on the table when there is an empty array in the list.

Thank you :)
Hoping to see the fixes soon,
Sorry for the late review.
Gui.

src/components/AccordionTable/AccordionTableBasic.part.tsx
```diff
@@ -234,17 +240,17 @@ export class AccordionTableBasic<T> extends React.Component<AccordionTableProps<
...
+    const { detailsRenderer, rowRenderer, theme, arrowToggle } = this.props;
     const { selectedIndex, expandedGroups } = this.state;
     const { groupedRows = [] } = state;
     const active = hasIndex(selectedIndex, index);
     const count = React.Children.count(cells);
     const col = this.getGroupByValue(data);
-    const open = !col || expandedGroups.indexOf(col) !== -1;
+    const open = !(col || col === 0) || expandedGroups.indexOf(col) !== -1;
     const renderData = { cells, index, data, active, key, state };
     // const isNewGroup = col && groupedRows.indexOf(col) === -1;
```
guiherzog: I would remove the comment here.

PR #138, opened 2019-07-30
IE11. Fix text input flex layout.

IKolosynskyi:
### Prerequisites

### Contribution Type

What types of changes does your code introduce?:

### Description

Text Box nested inside the flex container will have zero width in the IE. 
image
It's happening because of flex-basis: 0.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "precise-ui",
-  "version": "0.6.3",
+  "version": "0.6.4",
```
dimabory: No need for bumping up the version. We released nothing new after `0.6.1`. I would leave `0.6.2` and update CHANGELOG with respective changes.

IKolosynskyi: The previous version was 0.6.3, I would not prefer to downgrade version. Also, the current Pharos dependency version is 
image
therefore I prefer to bump version.

dimabory: yes, looks like you've introduced `0.6.3` as well 😏

dimabory: I suggest discussing this with main maintainers @kotielnikov @FlorianRappl @boromo

IKolosynskyi: The Semver Standart says: Bug fixes not affecting the API increment the patch version. 0.6.3 -> 0.6.4 is the patch version incrementing.

kotielnikov: 1) there is no need to increase the version, it's already higher than on `master`. 
2) For prerelease version (anything lower than 1.0.0) minor becomes major and patch becomes minor.

dimabory: > Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
@IKolosynskyi your fix is definitely not a BC

kotielnikov:
> @kotielnikov @IKolosynskyi I think we should decrease the version to `0.6.2` and update changelog accordingly.

Latest released version is 0.6.3, packages are published already https://www.npmjs.com/package/precise-ui?activeTab=versions. **Decreasing version will lead to conflicts.**

@IKolosynskyi just use 0.6.3

dimabory:
> > @kotielnikov @IKolosynskyi I think we should decrease the version to `0.6.2` and update changelog accordingly.
> 
> Latest released version is 0.6.3, packages are published already [npmjs.com/package/precise-ui?activeTab=versions](https://www.npmjs.com/package/precise-ui?activeTab=versions). **Decreasing version will lead to conflicts.**
> 
> @IKolosynskyi just use 0.6.3

Actually, there won't be any conflicts due to pre-release version. It will be just like another version `0.6.2-pre.xxx`.

kotielnikov:
> > > @kotielnikov @IKolosynskyi I think we should decrease the version to `0.6.2` and update changelog accordingly.
> > 
> > 
> > Latest released version is 0.6.3, packages are published already [npmjs.com/package/precise-ui?activeTab=versions](https://www.npmjs.com/package/precise-ui?activeTab=versions). **Decreasing version will lead to conflicts.**
> > @IKolosynskyi just use 0.6.3
> 
> Actually, there won't be any conflicts due to pre-release version. It will be just like another version `0.6.2-pre.xxx`.

Please fix me if I'm wrong.

More functionality -> higher version (this how versioning works). You can not go another way.

dimabory:
> Please fix me if I'm wrong.
> 
> More functionality -> higher version (this how versioning works). You can not go another way.

**SemVer applies to released packages** comprising the changes listed in `CHANGELOG`. Quoting you `More functionality -> higher version` I 'm pretty sure there is missed one stage, which is `releasing the new functionality`.
That's why the next version can be one of:
- `1.0.0` (ready for/used in production, stable API, or the new version contains _backwards incompatible changes_);
- `0.7.0` (there is _backwards compatible_ changes to the public API);
- `0.6.2` (_only backwards compatible bug fixes_).

Summarizing, we shouldn't increase the version each time when the PR merged into `develop`. It's the usual case when the PATCH release includes a lot of bugfixes.

IKolosynskyi:
@dimabory Version decreasing is possible only if we don't have version above, therefore I will not change the version to 0.6.2, this will bring mess into the current versioning system. Also, after this, my updated will not be the latest from the npm perspective.

kotielnikov:
> Summarizing, we shouldn't increase the version each time when the PR merged into develop. It's the usual case when the PATCH release includes a lot of bugfixes.

That's totally correct, no arguing about this. The problem is that version was increased already and package was published.###  There is no way back

dimabory:
> @dimabory Version decreasing is possible only if we don't have version above, therefore I will not change the version to 0.6.2, this will bring mess into the current versioning system. Also, after this, my updated will not be the latest from the npm perspective.

It shouldn't be the latest and even won't be. The latest is `0.6.1`Try to check it [here](https://semver.npmjs.com/) by picking a package `precise-ui` and entering the range `latest`.

IKolosynskyi:
> > @dimabory Version decreasing is possible only if we don't have version above, therefore I will not change the version to 0.6.2, this will bring mess into the current versioning system. Also, after this, my updated will not be the latest from the npm perspective.
> 
> It shouldn't be the latest and even won't be. The latest is `0.6.1`. Try to check it [here](https://semver.npmjs.com/) by picking a package `precise-ui` and entering the range `latest`.

image

Version decreasing will force to use the strict version.

kotielnikov:
@dimabory `latest` refers to the latest stable version. We are talking here about version numbers and what is the latest in sense of version number. @IKolosynskyi is totally right here - decreasing a version  will lead to the mess in versioning system

dimabory:
> > > @dimabory Version decreasing is possible only if we don't have version above, therefore I will not change the version to 0.6.2, this will bring mess into the current versioning system. Also, after this, my updated will not be the latest from the npm perspective.
> > 
> > 
> > It shouldn't be the latest and even won't be. The latest is `0.6.1`. Try to check it [here](https://semver.npmjs.com/) by picking a package `precise-ui` and entering the range `latest`.
> 
> image
> 
> Version decreasing will force to use the strict version.

Let's use it again https://semver.npmjs.com/.
- package: `precise-ui`
- version: `^0.6.2-pre.358`

What do you expect to see as a selected latest version?

IKolosynskyi:
> > > > @dimabory Version decreasing is possible only if we don't have version above, therefore I will not change the version to 0.6.2, this will bring mess into the current versioning system. Also, after this, my updated will not be the latest from the npm perspective.
> > > 
> > > 
> > > It shouldn't be the latest and even won't be. The latest is `0.6.1`. Try to check it [here](https://semver.npmjs.com/) by picking a package `precise-ui` and entering the range `latest`.
> > 
> > 
> > image
> > Version decreasing will force to use the strict version.
> 
> Let's use it again https://semver.npmjs.com/.
> 
> * package: `precise-ui`
> * version: `^0.6.2-pre.358`
> 
> What do you expect to see as a selected latest version?

I want to see the highest version but after your proposition, I will see the intermediate version.

dimabory:
> I want to see the highest version but after your proposition, I will see the intermediate version.

What do you mean under `intermediate version`? 
If you decrease the version now to `0.6.2` the next pre-release version like `0.6.2-pre.xxx` will contain everything from `^0.6.3-pre.xxx` as well. Afterwards, we have to update (decrease) the version in the dependencies and `npm i` do everything for you.

IKolosynskyi:
@dimabory Let's stop this versioning holy war. Two vs one said that it's better to live 0.6.3. I don't see the benefits of going back.

kotielnikov:
@dimabory 
Lets assume: 
I'm a person from outside, have no idea about our internal organization stuff. And I want to use `precise-ui`, a super-duper ui library. And I'm always into the new stuff. NEWEST STUFF. I just can not deal with the idea that my project is not using the most recent version. That's why I'm checking the [registry](https://www.npmjs.com/package/precise-ui?activeTab=versions) and see that the highest version is 0.6.3-pre.442. I install it.

Oh wait, but it doesn't contain the latest code, because someone messed up with versioning.

kotielnikov:
> @dimabory Let's stop this versioning holy war. Two vs one said that it's better to live 0.6.3. I don't see the benefits of going back. Sometimes standards != real life.

it's important that all team members understand and accept these rules

dimabory:
> > @dimabory Let's stop this versioning holy war. Two vs one said that it's better to live 0.6.3. I don't see the benefits of going back. Sometimes standards != real life.
> 
> it's important that all team members understand and accept these rules

@kotielnikov @IKolosynskyi I will approve it and hope we won't fall into messy with versions in the future. Thanks!

Issue #50, opened 2019-03-18
Padding left for Errors on InputComponent


emhaas:
### Description

The padding for the error message on InputComponents such as Textfields, Dropdowns etc. to the left side is set to 0, but instead it should be 1 rem.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

**Environment details:** [iOS, Chrome Version 73.0.3683.75 (Offizieller Build) (64-Bit), ...]

### Possible Solution

Current Implementation:
Padding 4px 0px
![image](https://user-images.githubusercontent.com/39331228/54533552-c1f6c400-498a-11e9-9e88-fdbbc9785694.png)

Expected Implementation:
Padding 4px 16px
![image](https://user-images.githubusercontent.com/39331228/54533570-ccb15900-498a-11e9-92d7-ec0bc1ee8cda.png)

PR #139, opened 2019-07-30
Fixed horizontal padding for InputInfo and InputError

ZEISSLaurinAgostini:
### Prerequisites

### Contribution Type

### Description

Fixed issue https://github.com/ZEISS/precise-ui/issues/50 with correct padding for info and error

**Before**
info_old
error_old
**Now**
info_new
error_new


emhaas:
For RadioButton and Checkbox the error text should be alligned with the main label text. 
Bildschirmfoto 2019-07-31 um 09 51 30
Bildschirmfoto 2019-07-31 um 09 50 28

dimabory:
@emhaas Please take a look at the component when there is no 'label'. Should it look like this? 
https://raw.githubusercontent.com/ZEISS/precise-ui/88737477381d40de416b59210ed14b94f76f59ba/integration/__image_snapshots__/TagBuilder_9-snap.png

ZEISSLaurinAgostini:
I would say yes. The horizontal alignment of all 3 elements (text, label, info/error) is always the same, even if just one of them is there.

emhaas:
As this isn't often or at all the case, it can be like that!

emhaas:
Text should be alligned with the big text but this is not a realistic use case anyway...

dimabory:
> Text should be alligned with the big text but this is not a realistic use case anyway...

Nothing impossible and we can (should) deal with it 😉

src/components/Checkbox/index.tsx
```diff
@@ -96,10 +77,12 @@ const RealCheckbox = styled.input`
   display: none;
 `;
 
-const FlexContainer = styled.div`
+const FlexContainer = styled.div.attrs(({ withInputInfo = false }: { withInputInfo: boolean }) => ({
```
kotielnikov: why are you using `attrs` here?

src/hoc/withInputInfo.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
```
kotielnikov: You have wrong understanding of High Order Components: High order component stands for the **enhancing** of other basic component. What are you doing here is a conditional rendering, which can be easily archived with a normal component. Please refactor it to `InputNotification` and add to the components folder.

src/components/PaddedContainer/index.tsx
```diff
@@ -32,10 +32,10 @@ interface PaddedContainerProps {
...
 const StyledContainer = styled('div')<PaddedContainerProps>(
   ({ top, left, bottom, right, gutter }) => css`
-    ${top ? `padding-top: ${distance[top]};` : ''}
-    ${left ? `padding-left: ${distance[left]};` : ''}
-    ${bottom ? `padding-bottom: ${distance[bottom]};` : ''}
-    ${right ? `padding-right: ${distance[right]};` : ''}
+    padding-top: ${(top && distance[top]) || 0};
+    padding-left: ${(left && distance[left]) || 0};
+    padding-bottom: ${(bottom && distance[bottom]) || 0};
+    padding-right: ${(right && distance[right]) || 0};
```
kotielnikov: why these changes are needed?

src/components/RadioButton/index.tsx
```diff
@@ -110,9 +111,11 @@ const SelectMark = styled('div')<RadioButtonCircleProps>`
   transform: ${props => (props.selected ? 'scale(1)' : 'scale(0)')};
 `;
 
-const FlexContainer = styled.div`
+const FlexContainer = styled.div.attrs(({ withError = false }: { withError: boolean }) => ({
```
kotielnikov: why are using `attrs`?

src/utils/input.tsx
```diff
@@ -1,21 +1,7 @@
 import * as React from 'react';
-import { InputError } from '../components/InputError';
-import { InputInfo } from '../components/InputInfo';
+import { withInputInfo } from '../hoc/withInputInfo';
 
 export function showInputInfo(error?: React.ReactChild, info?: React.ReactChild) {
```
kotielnikov: export this method from the new introduced component `InputNotification`

src/components/Checkbox/index.tsx
dimabory: why not?

src/components/PaddedContainer/index.tsx
dimabory: to forcibly set padding to `0px`

src/hoc/withInputInfo.tsx
dimabory: `InputNotification` cannot be a react component. 
Just take a look what's returning by `withInputInfo` (and I think it's kinda enhancing).

src/components/Checkbox/index.tsx
kotielnikov: Please read the documentation 
https://www.styled-components.com/docs/faqs#when-to-use-attrs

src/hoc/withInputInfo.tsx
kotielnikov: Please read carefully the documentation:
https://reactjs.org/docs/higher-order-components.html

> Concretely, a higher-order component is a function that takes a component and returns a new component.

It's called **enhancing** .

src/components/PaddedContainer/index.tsx
kotielnikov: Thanks, I know what `padding` css property means. The question was why do you need to set it forcibly?

src/hoc/withInputInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+      </InputError>
+    ) : (
+      <>{error}</>
+    );
+  };
+
+  if (error) return (props: InputErrorProps) => <WithError error={error} {...props} />;
+  if (info) return (props: InputInfoProps) => <WithInfo info={info} {...props} />;
+
+  return () => <></>;
```
kotielnikov: React.Fragments are for different purpose - do group several react nodes. If you don't have nodes, then just not use fragment. use **null**

src/hoc/withInputInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return typeof error === 'string' ? (
+      <InputError left="medium" {...props}>
+        {error}
+      </InputError>
+    ) : (
+      <>{error}</>
+    );
+  };
+
+  if (error) return (props: InputErrorProps) => <WithError error={error} {...props} />;
```
kotielnikov: if/else one-liners are agains our style-guides

src/hoc/withInputInfo.tsx
dimabory: Can we consider `React.ReactChild` as a component?

src/components/PaddedContainer/index.tsx
dimabory: To override default props if it is existing. For instance, take a look at https://github.com/ZEISS/precise-ui/pull/139/files?file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.tsx#diff-98434080865c9ea31f591a968fbe9909

By default `InputInfo` has `padding-left="medium"` and now it is possible to  forcibly reset this value.

src/hoc/withInputInfo.tsx
dimabory: can you provide me the link to the style-guide?

src/hoc/withInputInfo.tsx
kotielnikov: React component:
```
const Component = () => <div />
```

ReactChild, which is also a `ReactElement | string | number`:
```
<Component />
```

A good article describing principles:
https://medium.com/@fay_jai/react-elements-vs-react-components-vs-component-backing-instances-14d42729f62

kotielnikov:
```

export interface InputNotificationProps extends InputErrorProps, InputInfoProps {
  error?: React.ReactNode;
  info?: React.ReactNode;
}

export const InputNotification: React.FC<InputNotificationProps> = ({ error, info, ...rest }) => {
  if (error) {
    if (typeof error === 'string') {
      return <InputError {...rest}>{error}</InputError>;
    }

    return <>{error}</>;
  } else if (info) {
    if (typeof info === 'string') {
      return <InputInfo {...rest}>{info}</InputInfo>;
    }

    return <>{info}</>;
  }

  // tslint:disable-next-line
  return null;
};

/**
 * @deprecated use InputNotification component instead
 * @param error ReactChild
 * @param info ReactChild
 */
export function showInputInfo(error?: React.ReactChild, info?: React.ReactChild) {
  return <InputNotification error={error} info={info} />;
}
```

src/components/InputError/index.tsx
```diff
@@ -1,21 +1,13 @@
...
-  /**
-   * The content of input error (error to display).
-   */
-  children?: React.ReactNode;
-}
+export interface InputErrorProps extends PaddedContainerProps {}
 
-const StyledError = styled('div')<InputErrorProps>`
-  ${getFontStyle({ size: 'xSmall' })}
+const StyledError = styled(PaddedContainer)<InputErrorProps>`
```
kotielnikov: `styled.div`

I think there was a misunderstanding. Gutters should be up to the consumer. Semantically error/info notification style has nothing to do with `PaddedContainer`, while  `PaddedContainer` is a container and Info/Error Input is styling of error notifications. they serve different purposes. It has no sense that this component has the container props.

When you need to add gutters to InputNotification you will use composition either through styled-components api or just as a basic composition

i.e
```
// through styled-components
const NotificationWithGutter = styled(InputNotification)`
    padding: ${distance.small};
`
```
or
```
// standard composition
<PaddedContainer><InputNotification {....} /></PaddedContainer>
```

src/components/InputInfo/index.tsx
```diff
@@ -1,20 +1,13 @@
...
-  /**
-   * The content of input info (text to display).
-   */
-  children?: React.ReactNode;
-}
+export interface InputInfoProps extends PaddedContainerProps {}
 
-const StyledInfo = styled('div')<InputInfoProps>`
-  ${getFontStyle({ size: 'xSmall' })}
+const StyledInfo = styled(PaddedContainer)<InputInfoProps>`
```
kotielnikov: styled.div

PR #142, opened 2019-08-06
Feature/126 accordion table sort and group no value fields

lestard:
### Prerequisites

### Contribution Type

### Description

See #126. 

- fix sorting of items that don't have a value for the `sortBy`/`groupBy` column
- change order of items so that items without a value are moved to the end of the list
- `AccordionTable`: 
	- add a separate pseudo-group for items that don't have a value for the `groupBy` column
	- add an optional prop to define the label for this new pseudo-group
- change sorting for values of type number so that `0` isn't handled as `falsy` anymore but instead is sorted and grouped like other numbers

Breaking changes: At the moment the handling of Tables with items that don't have a value defined for the `sortBy`/`groupBy` column isn't working correctly so I assume that most people don't have not-defined values in their data. However, as this PR changes the way not-defined values are handled in theory it could change the behaviour of existing apps that depend on the (wrong) sorting and grouping of not-complete items.


src/components/AccordionTable/AccordionTable.types.part.tsx
```diff
@@ -108,6 +108,12 @@ export interface AccordionTableProps<T> extends TableProps<T>, AccordionCardLabe
    * Notification callback if a group is expanded or collapsed.
    */
   onToggleGroup?(e: AccordionGroupToggleEvent<T>): void;
+
+  /**
+   * A label to be used for the group of those items which have no value for the 'groupBy' column.
+   * Default value is `—`.
```
kotielnikov: @default '-'

src/components/AccordionTable/AccordionTableBasic.part.tsx
```diff
@@ -186,16 +220,40 @@ export class AccordionTableBasic<T> extends React.Component<AccordionTableProps<
     );
   }
 
+  private getGroupByValue = (rowData: any) => {
```
kotielnikov: why `any`?

lestard: If I use `T` instead I have the problem with the check `if (rowValue || rowValue === 0)`. TypeScript tells me that `rowValue` won't ever be a number and therefore the expression `rowValue === 0` will always be false.

src/components/AccordionTable/AccordionTable.types.part.tsx
kotielnikov: with "@": `@default: '-'`

lestard: I'm a little confused now on what the correct syntax is. Would this be correct:
```
  /**
   * A label to be used for the group of those items which have no value for the 'groupBy' column.
   * Default value is `—`.
   *
   * @default '—'
   */
```

kotielnikov:
```
  /**
   * A label to be used for the group of those items which have no value for the 'groupBy' column.
   *
   * @default '—'
   */
```

lestard:
Hi @dimabory. This is true, the first PR was merged to master by mistake and therefore the changes were not included in newer alpha-releases on npm. I talked to @guiherzog and he suggested to create a new PR targeting `develop` this time. 
Sorry for the trouble.

PR #156, opened 2019-08-14
Implement new width breakpoints and screen sizes

IKolosynskyi:
### Prerequisites

### Contribution Type

### Description

These changes will allow to more precisely display content for users with big monitor resolution.
New breakpoints:
  medium: 740, // tablet
  large: 980, // desktop
  xLarge: 1200, // hd desktop
  max: 1800, // full hd desktop

### Breaking changes

Earlier `large` was the biggest screen size and displayTo used to means: `(min-width: 'large' breakpoint) `**WITHOUT**` max-width breakpoint`.
Starting from this PR `large` screen size is not the biggest one, and `displayTo('large')` will generate  `(min-width: 'large' breakpoint and max-width: 'xLarge' breakpoint)`. 

To have the previous experience, in this PR were introduced new functions - `displayFrom, displayUpTo`. 
So old `displayTo('large')` will be equivalent to `displayFrom('large')`.


src/utils/displayTo.ts
```diff
@@ -1,33 +1,85 @@
...
+  }
+}
+
+export function formatQuery(previousBreakpoint?: number, nextBreakpoint?: number) {
+  if (previousBreakpoint == undefined && nextBreakpoint == undefined) {
+    throw new Error('Invaild breakpoints');
+  }
+
+  const queries = [];
+  if (previousBreakpoint != undefined) {
```
dimabory: please add strict comparison  `previousBreakpoint !== undefined`

src/utils/displayTo.ts
```diff
@@ -1,33 +1,85 @@
...
+    queries.push(`(min-width: ${previousBreakpoint}px)`);
+  }
+  if (nextBreakpoint != undefined) {
```
dimabory: please add strict comparison  `nextBreakpoint !== undefined`

src/utils/displayTo.ts
```diff
@@ -1,33 +1,85 @@
...
+    case 'xLarge':
+      return { prev: breakpoints.xLarge, next: breakpoints.max };
+    case 'max':
+      return { prev: breakpoints.max };
+    default:
+      return undefined;
+  }
+}
+
+export function formatQuery(previousBreakpoint?: number, nextBreakpoint?: number) {
```
dimabory: I suggest changing arguments to object type `{ prev?: number, next?: number }`

src/utils/displayTo.ts
```diff
@@ -1,33 +1,85 @@
...
+  if (previousBreakpoint == undefined && nextBreakpoint == undefined) {
```
dimabory: why you don't use strict comparison again?

src/themes.ts
```diff
@@ -26,6 +26,8 @@ const colorCycle = [
 export const breakpoints: Breakpoints = {
```
dimabory: why there is no `'small'` breakpoint?

IKolosynskyi: `small` is not breakpoint, `small` is the screen size from 0 to medium breakpoint.

src/utils/displayTo.ts
IKolosynskyi: In the current case, I prefer to not use an object as an argument.

dimabory: In that case, I should always remeber what the first parameter means when calling `formatQuery(undefined, 980)`.

IMO: `formatQuery({ next: 980 }}` loosk better.

src/components/Responsive/Example.md
```diff
@@ -34,27 +34,26 @@ const { Responsive } = require('precise-ui');
 ```
 
 **Utility Functions**
-
-Utils also exports `displayTo[screenSize]` function for styled components. Where the screen size is `small | smallAndMedium | medium | mediumAndLarge | large`.
+Utils also exports `displayTo[ScreenSize | query], displayUpTo(ScreenSize), displayFrom(ScreenSize)` function for styled components.
```
dimabory: should be `displayTo(ScreenSize | query)`

src/utils/displayTo.ts
IKolosynskyi: Got your point, you`re right. I will update this

src/utils/displayTo.ts
```diff
@@ -1,33 +1,86 @@
...
+      return { prev: breakpoints.xLarge, next: breakpoints.max };
+    case 'max':
+      return { prev: breakpoints.max };
+    default:
+      return undefined;
+  }
+}
+
+// create media query based on previous and next breakpoints
+export function formatQuery(breakpoints: { prev?: number; next?: number }) {
```
dimabory: You can skip 73 line
```suggestion
export function formatQuery({ next, prev }: { prev?: number; next?: number }) {
```

IKolosynskyi: Let the name of the variable `breakpoint` be visible

src/utils/displayTo.ts
```diff
@@ -1,33 +1,86 @@
...
+  const { next, prev } = breakpoints;
+  if (prev === undefined && next === undefined) {
+    throw new Error('Invaild breakpoints');
```
dimabory:
```suggestion
    throw new Error('Both breakpoints cannot be `undefined`');
```

src/common.ts
```diff
@@ -490,7 +492,10 @@ export interface RefProps {
   innerRef?(node: HTMLElement | null): void;
 }
 
-export type ScreenSize = 'small' | 'smallAndMedium' | 'medium' | 'mediumAndLarge' | 'large';
+export type ScreenSize = 'small' | 'smallAndMedium' | 'medium' | 'mediumAndLarge' | 'large' | 'xLarge' | 'max';
```
dimabory:
```suggestion
/** @deprecated */
export type CombinedScreenSize = 'smallAndMedium' | 'mediumAndLarge';
export type ScreenSize = 'small' | 'medium' | 'large' | 'xLarge' | 'max' | CombinedScreenSize;
```

src/utils/displayTo.ts
```diff
@@ -1,33 +1,88 @@
...
+export function getScreenSizeBreakpoints(
+  screen: ScreenSize,
+  breakpoints: Breakpoints,
+): { next?: number; prev?: number } | undefined {
+  switch (screen) {
+    case 'small':
+      return { next: breakpoints.medium };
+    // `smallAndMedium` deprecated and will be deleted in the future
+    case 'smallAndMedium':
+      return { next: breakpoints.large };
```
dimabory:
```suggestion
      console.warn("'smallAndMedium' screen size has been deprecated since 0.8.0.");
      return { next: breakpoints.large };
```

src/utils/displayTo.ts
```diff
@@ -1,33 +1,88 @@
...
+    case 'medium':
+      return { prev: breakpoints.medium, next: breakpoints.large };
+    // `mediumAndLarge` deprecated and will be deleted in the future
+    case 'mediumAndLarge':
+      return { prev: breakpoints.medium };
```
dimabory:
```suggestion
      console.warn("'mediumAndLarge' screen size has been deprecated since 0.8.0.");
      return { prev: breakpoints.medium };
```

src/utils/displayTo.ts
IKolosynskyi: Console.warn() was added.

src/utils/displayTo.ts
```diff
@@ -1,33 +1,88 @@
...
+export function displayFrom(screen: ScreenSize) {
+  const screenBreakpoints = getScreenSizeBreakpoints(screen, breakpoints);
+  if (!screenBreakpoints) {
+    throw new Error('Invalid screen size');
+  }
+  const query = formatQuery({ prev: screenBreakpoints.prev });
+  return formatMedia(query);
 }
 
 export function displayTo(screen: ScreenSize | string): ThemedCssFunction<any> {
```
dimabory: Are we going to deprecate this function as well?

IKolosynskyi: No

src/utils/displayTo.ts
```diff
@@ -1,33 +1,88 @@
...
+  switch (screen) {
+    case 'small':
+      return { next: breakpoints.medium };
+    case 'smallAndMedium':
+      console.warn("'smallAndMedium' screen size has been deprecated since 0.8.0. Use 'displayUpTo' function.");
+      return { next: breakpoints.large };
+    case 'medium':
+      return { prev: breakpoints.medium, next: breakpoints.large };
+    case 'mediumAndLarge':
+      console.warn("'mediumAndLarge' screen size has been deprecated since 0.8.0. Use 'displayFrom' function.");
```
dimabory: `Use 'displayFrom/displayUpTo' function.`. I'm not sure if it's necessary. Please remove.

IKolosynskyi: I think it's necessary because when you say that something is deprecated you should provide an alternative way.

IKolosynskyi: In addition, extra-information has never been a problem.

dimabory: Agree 😀, but see an example:

```ts
displayFrom('mediumAndLarge');
```
then console warning will show me the warning:
```bash
"'mediumAndLarge' screen size has been deprecated since 0.8.0. Use 'displayFrom' function."
```

Do you think it's a developer-friendly error message if I've already used `displayFrom` function?

IKolosynskyi: Updated message

IKolosynskyi:
> can we have custom screen size for newly created functions as well as `displayTo` has?
> 
> `displayUpTo('70em')`
> `displayFrom('777px')`

It's bad practice to use custom and points. But if you really want, you can do something like below

`
var query = formatQuery({next/prev: 700});
displayTo(query);

OR
I`ve never head of using em/rem for the media query, but

displayTo('(min-width: 70em)');

`

dimabory:
> It's bad practice to use custom and points. But if you really want, you can do something like below
> 

I don't mind to not use custom media queries, but if `displayTo` still has the option to do this, therefore I think the newly introduced functions (which does almost the same) should support it as well. 

> I`ve never heard of using em/rem for the media query

If it's possible then somebody can use it.

IKolosynskyi:
> > It's bad practice to use custom and points. But if you really want, you can do something like below
> 
> I don't mind to not use custom media queries, but if `displayTo` still has the option to do this, therefore I think the newly introduced functions (which does almost the same) should support it as well.
`displayTo` doesn't support using 800px or 88em as an argument.
> 
> > I`ve never heard of using em/rem for the media query
> 
> If it's possible then somebody can use it.

They still can manually formate query and pass it to displayTo

dimabory:
> `displayTo` doesn't support using 800px or 88em as an argument.

Yeah, sure, I mean `displayTo('(min-width: 800px)')`

> They still can manually formate query and pass it to displayTo

Perfect. Can you add an example of how to use `formatQuery`?

IKolosynskyi:
> > `displayTo` doesn't support using 800px or 88em as an argument.
> 
> Yeah, sure, I mean `displayTo('(min-width: 800px)')`
if `displayTo` doesn't support `%num%px` argument then `displayFrom/UpTo` doesn't have to support this as well.
> 
> > They still can manually formate query and pass it to displayTo
> 
> Perfect. Can you add an example of how to use `formatQuery`?
Yes, I will add this.

dimabory:
> Yes, I will add this.

Thanks a lot!

src/utils/displayTo.ts
```diff
@@ -1,33 +1,92 @@
...
 export function displayTo(screen: ScreenSize | string): ThemedCssFunction<any> {
-  const query = getMediaQueries(breakpoints)[screen] || screen;
-  const result: ThemedCssFunction<any> = (strings: any, ...interpolations: Array<any>) =>
+  const screenBreakpoints = getScreenSizeBreakpoints(screen as ScreenSize, breakpoints);
+  const query = screenBreakpoints ? formatQuery(screenBreakpoints) : screen;
+  return formatMedia(query);
+}
+
+// create media query based on previous and next breakpoints
+export function formatQuery(breakpoints: { prev?: number; next?: number }) {
```
dimabory: Btw, if it's a public API, the `breakpoints` arg should be `{ min: number, max: number }` to be more readable as well as developer-friendly.
Just compare:
`displayTo(formatQuery({ prev: 300, next: 600}))` vs `displayTo(formatQuery({ min: 300, max: 600}))`

And last but not least you should validate whether `min < max` if both breakpoints provided.

dimabory:
```suggestion
export function formatMediaQuery(breakpoints: { prev?: number; next?: number }) {
```

dimabory:
> > > `displayTo` doesn't support using 800px or 88em as an argument.
> > 
> > 
> > Yeah, sure, I mean `displayTo('(min-width: 800px)')`
> 
> if `displayTo` doesn't support `%num%px` argument then `displayFrom/UpTo` doesn't have to support this as well.

`displayTo` supports `displayTo('min-width: 200px')`

src/utils/displayTo.ts
IKolosynskyi: Updated

IKolosynskyi: Added formatQuery example

src/components/Responsive/Example.md
```diff
@@ -74,3 +78,15 @@ const StyledDiv6 = styled.div`
   <StyledDiv6>It's bold on screens between 200px and 500px.</StyledDiv6>
 </>;
 ```
+
+In addition, utils exports `getWidthBreakpointsQuery({max, min})`.
+The function `getWidthBreakpointsQuery` formates width media query.
+
+```jsx
+const styled = require('styled-components').default;
```
dimabory:
```suggestion
const { getWidthBreakpointsQuery, styled } = require('precise-ui');
```

src/utils/displayTo.ts
```diff
@@ -1,33 +1,93 @@
...
+    case 'xLarge':
+      return { min: breakpoints.xLarge, max: breakpoints.max };
+    case 'max':
+      return { min: breakpoints.max };
+    default:
+      return undefined;
+  }
+}
+
+function getMediaQueryFunc(query: string): ThemedCssFunction<any> {
```
dimabory:
```suggestion
function getMediaQuery(query: string): ThemedCssFunction<any> {
```

IKolosynskyi: getMediaQueryFunc return func, not media query

dimabory: It's an internal function, that's why I don't see the reason for `Func` suffix

package.json
```diff
@@ -20,7 +20,6 @@
     "build:demo:test": "cross-env NODE_ENV=test styleguidist build",
     "build": "rimraf dist && npm run build:bundle && npm run build:es5 && npm run build:es6",
     "icongen": "node tools/convert-icons.js",
```
dimabory:
```suggestion
    "icongen": "node tools/convert-icons.js",
    "prepush": "npm run test & npm run build",
```

package.json
```diff
@@ -20,7 +20,6 @@
...
-    "prepush": "npm run test & npm run build",
```
dimabory: 👎

Issue #144, opened 2019-08-07
Incorrect position of label ellipsis


ZEISSLaurinAgostini:
### Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository?
- [x] Did you perform a search in the issues?

### Versions

0.6.3

### Description

![label_ellipsis](https://user-images.githubusercontent.com/47562891/62614402-9d848f80-b90b-11e9-8556-ae72452886b7.png)
![label_ellipsis2](https://user-images.githubusercontent.com/47562891/62614404-9d848f80-b90b-11e9-96af-da1ecd33557a.png)

### Steps to Reproduce

1. Create any Component with a set label property (e.g. TextField) or use the examples in the style guide
2. Decrease the width until the ellipsis appears

**Expected behavior:** Ellipsis should only appear if the label is too wide

**Actual behavior:** Ellipsis appears before the label is cut

PR #157, opened 2019-08-14
Bugfix/144 ellipsis position

ZEISSLaurinAgostini:
### Prerequisites

### Contribution Type

### Description

Fix for https://github.com/ZEISS/precise-ui/issues/144 
Setting `display` of placeholder text to `none` instead of just making it transparent on focus.


src/quarks/StyledInputRow.tsx
```diff
@@ -49,7 +49,7 @@ const TextFieldLabelText = styled('span')<TextFieldLabelProps>`
   background: ${themed(({ multiline, theme }) => (multiline ? theme.ui2 : 'transparent'))};
 
   > span {
-    opacity: ${props => (props.active || props.filled ? '0' : '1')};
+    display: ${props => (props.active || props.filled ? 'none' : '')};
```
dimabory: I'd prefer to use `visibility` in this case.

Can you try instead of `display` to use `visibility: hidden|visible`?

ZEISSLaurinAgostini: Using `visibility` won't work, because we need to make sure the inner `span` isn't used for calculating the width of the label.
I tried it with `visibility: hidden`, but the only thing working as intended was `display: none`.

package.json
```diff
@@ -38,6 +38,8 @@
   "author": "ZEISS Digital Innovation Partners",
   "license": "MIT",
   "dependencies": {
+    "date-fns": "^2.0.0-beta.1",
+    "jest-environment-node": "^24.8.0",
```
dimabory: why this deps here?

src/quarks/StyledInputRow.tsx
dimabory: Gotcha. Thanks for the explanation.

package.json
ZEISSLaurinAgostini: `date-fns` was there in 0.7.0 and removed in 0.7.1...looks like a merge issue.
I needed `jest-environment-node` for test:visual, but I didn't install it with `--save`, so I'm a bit surprised it's there.

dimabory: Please remove them

ZEISSLaurinAgostini: Travis CI build fails because of:
`Error: Cannot find module 'jest-environment-node'`

dimabory: it's weird because it has been working before

dimabory: then let's add `jest-environment-node`, but please make sure it is in `devDependencies` and not in `dependencies`

dimabory: Well, I see why it has been broken. You've committed `package-lock.json` with new/updated versions of dependencies.

ZEISSLaurinAgostini: So what can I do to fix it?

dimabory: Let's keep it in the deps list...

PR #161, opened 2019-08-20
Fix issue with AccordionTable and groupBy array fields

lestard:
### Prerequisites

### Contribution Type

### Description

With #142 the handling of `groupBy` was changed and an issue with columns of type array was introduced. This PR fixes an issue when the column that is grouped by is an array.

Before:
- the counting of items in the group was wrong
- some items were missing in the groups
- in the group-label the items of the array were concatenated without any separator. So if the column value is `["A", "B", "C"]` then the label was `"ABC"` while the desired value would be `"A,B,C"` which is what `toString` is creating for arrays.

The root cause of those issues was that if you have an array of arrays you cannot simply test if an array is included by using `indexOf(item) !== -1` because strict equality check is not working for arrays.


guiherzog:
Hi @lestard, can you please update your branch? It's out of sync with the current develop.

PR #180, opened 2019-10-07
Fix Datepicker bugs: incorrect example date format, controlled mode issues

IKolosynskyi:
### Prerequisites

### Contribution Type

### Description

[Place a meaningful description here.]


src/utils/date.ts
```diff
@@ -76,6 +76,25 @@ export function parseDate(
   return isValid(parsedDate) && strictParsingValueMatch ? parsedDate : null;
 }
 
+export function getDefaultDateFormat() {
+  const formatObj = new Intl.DateTimeFormat().formatToParts(new Date());
```
dimabory: Did you test it in IE11?

src/components/DateField/Example.md
```diff
@@ -343,7 +343,7 @@ In `value` controlled mode
 const { DateField, StackPanel, StackItem, Button } = require('precise-ui');
 
 const Example = () => {
-  const [ value, setValue ] = React.useState('14-06-1988');
+  const [ value, setValue ] = React.useState(new Date());
```
dimabory: Are you sure this is valid?

`DateFieldProps` type says it can accept `value` with type of `string|undefined`.

Please fix me if I'm not right.

src/utils/date.ts
IKolosynskyi: https://caniuse.com/#search=DateTimeFormat says it has to work but I will double-check it.

src/components/DateField/Example.md
IKolosynskyi: Yes but all Zeisslets that I've seen transmit Date and it works correctly. I suppose types should be changed to `string|Date|undefined` but it affects most of the code, so I prefer to leave it as it is.

src/utils/date.ts
dimabory: right, but not `formatToParts` https://caniuse.com/#search=formatToParts

src/components/DateField/Example.md
dimabory: 🤔 Can't find any example of using the `Date` instance instead of `string`...

IKolosynskyi: Example reworked

Issue #123, opened 2019-07-16
Selecting element from autosuggestion IE


MitaliDhamale:
**Description:**
Selecting an element from autosuggestion do not close the suggestions.

**Steps to reproduce:**

// todo

**Expected behavior:**
The autosuggestion should be closed once the user selects an element from autosuggestion

**Actual behavior:**
The selected element is still displayed as a suggestion to the user in IE

**Environment Details:**
 IE11

dimabory:
@MitaliDhamale Can you use the default template for bug's description https://github.com/ZEISS/precise-ui/blob/master/.github/ISSUE_TEMPLATE/bugs.md?

PR #184, opened 2019-10-10
fix(#123): Selecting element from autosuggestion IE

ZEISSLaurinAgostini:
### Prerequisites

### Contribution Type

### Description

This solves #123 

On IE11 the focus event from `this._element.focus();` was triggered after `this.hide()` and so the suggestion list opened again after hiding it.

Adding an additional call to `this.hide()` after a 10ms delay solves this. It's a bit hacky, but I didn't find a better solution.


ZEISSLaurinAgostini:
It looks like the build stalled. Something with building the Docker image. Does someone know why it takes so long?

ZEISSLaurinAgostini:
@dimabory Is there a way to restart the build without adding a new (unnecessary) commit?

dimabory:
> @dimabory Is there a way to restart the build without adding a new (unnecessary) commit?

Sure, hope you have got a permission
image

src/components/Autocomplete/index.tsx
```diff
@@ -181,7 +181,16 @@ class AutocompleteInt<T> extends React.Component<AutocompleteProps<T> & FormCont
...
     }
 
-    suggestionSelected ? this.hide() : this.show();
+    const hide = () => {
+      this.hide();
+      setTimeout(() => {
+        if (this.state.open) {
+          this.hide();
+        }
+      }, 10);
```
dimabory: why `10`? Can you try to omit the second param? I guess it can also make it work.

src/components/Autocomplete/index.tsx
```diff
@@ -181,7 +181,16 @@ class AutocompleteInt<T> extends React.Component<AutocompleteProps<T> & FormCont
...
     }
 
-    suggestionSelected ? this.hide() : this.show();
+    const hide = () => {
+      this.hide();
+      setTimeout(() => {
+        if (this.state.open) {
+          this.hide();
+        }
+      });
```
dimabory: Could you please add the comment explaining this hack😊?

PR #187, opened 2019-10-25
Feature/185 add onsort event to the table

herrkris:
### Prerequisites

### Contribution Type

### Description

When passing a `onSort` callback property to the Table every time the user sorts the table the passed callback gets executed with the `columnKey` and the `order` value or `undefined` based on the current sorting state.

### DIFF
--- a/src/components/Table/Example.md
+++ b/src/components/Table/Example.md
@@ -422,7 +422,7 @@ const data = [{
 
 Providing an `onSort` callback allows you to do additional data fetching if the user invokes sorting on your table.
 
-```jsx
+```jsx { "props": { "data-skip": true } }
 const { Table } = require('precise-ui');
 const data = [{
   name: 'A',
--- a/src/components/Table/Example.md
+++ b/src/components/Table/Example.md
@@ -438,5 +438,5 @@ const data = [{
   age: 15,
 }];
 
-<Table data={data} sortBy="age" onSort={(ev) => console.log(ev)} />
+<Table data={data} onSort={(ev) => console.log(ev)} />
 ```
--- a/src/components/Table/Example.md
+++ b/src/components/Table/Example.md
@@ -417,3 +417,26 @@ const data = [{
 
 <Table data={data} groupBy="name" />
 ```
+
+** onSort Callback **
+
+Providing an `onSort` callback allows you to do additional data fetching if the user invokes sorting on your table.
+
+```jsx
+const { Table } = require('precise-ui');
+const data = [{
+  name: 'A',
+  age: 21,
+}, {
+  name: 'B',
+  age: 12,
+}, {
+  name: 'C',
+  age: 25,
+}, {
+  name: 'D',
+  age: 15,
+}];
+
+<Table data={data} onSort={(ev) => console.log(ev)} />
+```
--- a/src/components/Table/Example.md
+++ b/src/components/Table/Example.md
@@ -438,5 +438,5 @@ const data = [{
   age: 15,
 }];
 
-<Table data={data} onSort={(ev) => console.log(ev)} />
+<Table data={data} sortBy="age" onSort={(ev) => console.log(ev)} />
 ```
--- a/src/components/Table/Example.md
+++ b/src/components/Table/Example.md
@@ -420,7 +420,7 @@ const data = [{
 
 ** onSort Callback **
 
-Providing an `onSort` callback allows you to do additional data fetching if the user invokes sorting on your table.
+Providing an `onSort` callback allows you to react to the table being sorted.
 
 ```jsx { "props": { "data-skip": true } }
 const { Table } = require('precise-ui');
--- a/src/components/Table/Table.types.part.tsx
+++ b/src/components/Table/Table.types.part.tsx
@@ -45,7 +45,7 @@ export interface TableSortEvent {
    * The direction in which the table is sorted.
    * `undefined` represents the default unsorted state.
    */
-  order: 'ascending' | 'descending' | undefined;
+  order?: 'ascending' | 'descending';
 }
 
 /**
--- a/src/components/Table/Table.types.part.tsx
+++ b/src/components/Table/Table.types.part.tsx
@@ -186,6 +186,10 @@ export interface TableProps<T> extends StandardProps, ModeProviderProps<TableMod
    */
   onDataClick?(e: TableCellEvent<T>): void;
   /**
+   * Handler that is being called after the table has been sorted.
+   */
+  onSort?(e: TableCellEvent<T>): void;
+  /**
    * Optionally provides a custom way for computing the row key.
    */
   getRowKey?(e: TableRowEvent<T>): string;
--- a/src/components/Table/Table.types.part.tsx
+++ b/src/components/Table/Table.types.part.tsx
@@ -186,7 +186,10 @@ export interface TableProps<T> extends StandardProps, ModeProviderProps<TableMod
    */
   onDataClick?(e: TableCellEvent<T>): void;
   /**
-   * Handler that is being called after the table has been sorted.
+   * Handler that is being called after the table has internally been sorted.
+   * When the `onHeaderClick` or the `sortBy` property is present `onSort`
+   * will never execute. This handler only gets called when internal sorting
+   * is being used.
    */
   onSort?(e: TableCellEvent<T>): void;
   /**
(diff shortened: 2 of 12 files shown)


herrkris:
See #187 for more details

src/components/Table/Table.types.part.tsx
```diff
@@ -185,6 +185,10 @@ export interface TableProps<T> extends StandardProps, ModeProviderProps<TableMod
    * Handler that is being called when a data cell has been clicked.
    */
   onDataClick?(e: TableCellEvent<T>): void;
+  /**
+   * Handler that is being called after the table has been sorted.
+   */
+  onSort?(e: TableCellEvent<T>): void;
```
dimabory: It would be great to have a dedicated type - `TableSortEvent`. I don't think it's a good idea to re-use the same event-type for completely different events.

herrkris: I though about that as well and then saw that every table event uses the same `TableCellEvent` interface so I figured it would be desired behavior to have the same event interface here as well. But I'm happy to create a custom event for it. :)

src/components/Table/Example.md
```diff
@@ -417,3 +417,26 @@ const data = [{
...
+  age: 12,
+}, {
+  name: 'C',
+  age: 25,
+}, {
+  name: 'D',
+  age: 15,
+}];
+
+<Table data={data} onSort={(ev) => console.log(ev)} />
```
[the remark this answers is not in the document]
herrkris: Yeah, because the whole sorting is disable when `onHeaderClick` is present.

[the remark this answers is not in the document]
dimabory: Did you figure out the possible way to control the table sorting with `onHeaderClick` and `sortBy` props? 

I'm not 100% sure, but I guess it's possible to do it using them.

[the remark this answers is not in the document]
dimabory: Rough example for you case can be:
- pass initial `data={data}` to the component
- add event handler `onHeaderClick={fetchData}` (`fetchData = (sortBy?: SortingType) { ... }`)
- change value of `sortBy={sortBy}` within fetchData function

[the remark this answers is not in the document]
herrkris: That's definitely possible and I gave it a try. In the end I was rebuilding most of the default `TableBasic.part.tsx` internals and figured it would make more sense to be able to pass a `onSort` callback into the table.

src/components/Table/Table.types.part.tsx
herrkris: Any other thoughts on this @guiherzog or @boromo?

src/components/Table/Example.md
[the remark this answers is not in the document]
dimabory: > Yeah, because the whole sorting is disabled when `onHeaderClick` is present.
It would be good to have it documented(as a _JSDoc_) for `onSort` type.

src/components/Table/TableBasic.part.tsx
```diff
@@ -154,17 +154,24 @@ export class TableBasic<T> extends React.Component<TableProps<T> & RefProps, Tab
...
-      } else {
-        this.setState({
-          sorting: undefined,
+        newSortingValue = {
+          columnKey: key,
+          order,
+        };
+      }
+
+      this.setState({ sorting: newSortingValue });
```
dimabory: I suggest triggering `onSort` function inside `callback` argument of `setState`:
```ts
this.setState({ sorting: newSortingValue }, () => {
  if (typeof onSort === 'function') {
    onSort({
      column,
      key,
      value: newSortingValue && newSortingValue.order,
      row: -1,
    });
  }
});
```

dimabory: Same here https://github.com/ZEISS/precise-ui/pull/187/files#diff-0313273806b6ac65b35d01494c63ab75R214

src/components/Table/Example.md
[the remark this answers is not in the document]
herrkris: > Rough example for you case can be:
> 
> * pass initial `data={data}` to the component
> * add event handler `onHeaderClick={fetchData(sortBy)}` (`fetchData = (sortBy?: SortingType) { ... }`)
> * change value of `sortBy={sortBy}` inside _fetchData_ function

This would require to reimplement the whole [`getDefaultHeaderCellRenderer`](https://github.com/ZEISS/precise-ui/blob/86be817ae4e6ca08a996f941309126851c79d0b1/src/components/Table/TableShared.part.tsx#L62) on top of the sorting in the host component which I wanted to avoid.

src/components/Table/TableBasic.part.tsx
herrkris: Updated it accordingly

src/components/Table/Table.types.part.tsx
herrkris: I've added now a custom `TableSortEvent`.

src/components/Table/TableBasic.part.tsx
```diff
@@ -152,21 +152,31 @@ export class TableBasic<T> extends React.Component<TableProps<T> & RefProps, Tab
...
+          }
+
+          return { sorting: newSortingValue };
+        },
+        () => {
+          if (typeof onSort === 'function') {
+            onSort({
+              column,
+              key,
+              order: this.state.sorting && this.state.sorting.order,
```
guiherzog: You are destructuring the state on the param already, why do you need to call this.state again here?

herrkris: Because this is a differnt closure which has not access to the setState closure

guiherzog: Oops, my bad. I didn't see it was the callback of the setState. :)

src/components/Table/Table.types.part.tsx
```diff
@@ -29,6 +29,22 @@ export interface TableCellEvent<T> {
   data?: T;
 }
 
+export interface TableSortEvent {
```
dimabory: You have to export this type in [`src/components/Table/index.tsx`](https://github.com/ZEISS/precise-ui/blob/86be817ae4e6ca08a996f941309126851c79d0b1/src/components/Table/index.tsx#L14)

src/components/Table/Example.md
```diff
@@ -417,3 +417,26 @@ const data = [{
 
 <Table data={data} groupBy="name" />
 ```
+
+** onSort Callback **
+
+Providing an `onSort` callback allows you to do additional data fetching if the user invokes sorting on your table.
```
dimabory: Not sure if this description refers to the actual example provided below.  🤔

src/components/Table/Table.types.part.tsx
dimabory:
```suggestion
/**
 * Arguments for sorting events.
 */
export interface TableSortEvent {
```

src/components/Table/Table.types.part.tsx
```diff
@@ -29,6 +29,25 @@ export interface TableCellEvent<T> {
...
+  column: number;
+  /**
+   * The column key name that's sorted.
+   */
+  key: string;
+  /**
+   * The direction in which the table is sorted.
+   * `undefined` represents the default unsorted state.
+   */
+  order: 'ascending' | 'descending' | undefined;
```
dimabory:
```suggestion
  order?: 'ascending' | 'descending';
```

src/components/Table/Example.md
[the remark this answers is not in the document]
dimabory: Sounds like a headache 😟

src/components/Table/Table.types.part.tsx
herrkris: Both would work and are acceptable right? So I would go with the more explicit style of writing out `undefined`.

src/components/Table/Example.md
herrkris: Updated the description to be more generic

src/components/Table/Table.types.part.tsx
dimabory: Sure, I noticed a very similar type in [`TableSorting`](https://github.com/ZEISS/precise-ui/pull/187/files#diff-f3246e6f13ea56211990d0d9f195fcc9R275) and probably it makes sense to have it consistent. It's up to you.

dimabory:
LGTM  🤗 Thank you @herrkris  👍

PR #202, opened 2019-11-14
feat(Highlight): New Highlight Component

bajcmartinez:
### Prerequisites

- [X] I have read the **CONTRIBUTING** document
- [X] My code follows the code style of this project

### Contribution Type

- [X] New feature (non-breaking change which adds functionality, make sure to open an associated issue first)
- [X] My change requires a change to the documentation
- [X] I have updated the documentation accordingly
- [X] I have added tests to cover my changes
- [X] All new and existing tests passed

### Description

This new component allows developers to highlight text based on a term
condition


src/components/Hightlight/Example.md
```diff
@@ -0,0 +1,25 @@
+**Elementary**
```
dimabory: Could you also provide an example of how to change the color of a highlighted text?

src/components/Hightlight/index.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import styled, { themed } from '../../utils/styled';
+import { StandardProps } from '../../common';
+
+export interface HighlightProps extends StandardProps {
```
dimabory: I would also add the prop to change the color of a highlighted text and do not rely on global theme variables.

src/components/Hightlight/Example.md
bajcmartinez: Working on it!

src/components/Hightlight/index.tsx
bajcmartinez: Would it be ok to do something like: `theme={{highlightColor: '#f00'}}`?

dimabory: Fine, but as I mentioned before I don't like an idea to use **global** `theme` for very specific parts. The global theme should contain something that can be used across multiple components. For instance `theme.primary[Background]Color`, which can be used as (background) color of a button, header, footer, link, whatever else.

bajcmartinez: image

That would be an example on how to do it

dimabory: PS: I don't know which thoughts other contributors have regarding this topic, therefore I can't force you changing it now.

@boromo @IKolosynskyi @guiherzog

src/components/Highlight/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ignoreCase?: boolean;
+}
+
+const Highlighted = styled.span`
+  background-color: ${themed(props => props.theme.highlightColor)};
+`;
+
+/**
+ * Highlight component is meant to be a simple component to display text with highlighted search.
+ * Component will render a SPAN or series of SPAN with the content and highlights
```
dimabory: a ~~SPAN or~~ series of SPAN

src/components/Hightlight/index.tsx
bajcmartinez: I took the idea from other components, like the Accordion which does this same thing, just wanted to keep standard. Additionally, the color should be set on a global level to make sure in our implementation we have a standard look across different products.

src/components/Highlight/index.tsx
bajcmartinez: If there is no match it renders a single span

src/components/Hightlight/index.tsx
boromo: I believe that it makes sense to have it as theme color since in case one wants to change the theme, he would probably like to change the highlight as well.

src/components/Highlight/Example.md
```diff
@@ -0,0 +1,36 @@
...
+const { Highlight } = require('precise-ui');
+
+<Highlight text="Hello World!" highlight="World" />
+```
+
+**With exact match on case**
+
+`IgnoreCase` prop can be used to specify with the search will be case sensitive. By default `ignoreCase` is `True`.
+
+```jsx { "props": { "data-wait": 500 } }
```
dimabory: is `{ "data-wait": 500 }` really necessary for all examples?

src/components/Highlight/index.tsx
dimabory: really? as far as I can see the root is React.Fragment (which is not a span) and it contains a  series of spans.

dimabory: gotcha :)

src/components/Highlight/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const Highlight: React.FC<HighlightProps> = ({ text, highlight, ignoreCase = true, theme }) => {
```
dimabory: One small point:
I would immediately return the text in case of `highlight` prop equals an empty string (`""`).

src/components/Highlight/index.tsx
```diff
@@ -26,6 +26,10 @@ const Highlighted = styled.span`
  * Component will render a SPAN or series of SPAN with the content and highlights
  */
 export const Highlight: React.FC<HighlightProps> = ({ text, highlight, ignoreCase = true, theme }) => {
+  if (highlight === '') {
+    return <span>{text}</span>;
```
dimabory: why not just `return text;`?

bajcmartinez: Types, I'm retuning react elements and should be consistent, so can't just return a String here.

dimabory: Right, wouldn't be this better instead of wrapping by span `return <>{text}</>`;?

dimabory: Unfortunately, that's [the issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20544) of typings. React actually allows returning the string.

guiherzog:
@bajcmartinez could you sign the license/cla? 

@dimabory Are you finished with your review?

bajcmartinez:
I did already the other day: 
image

dimabory:
> @bajcmartinez could you sign the license/cla?
> 
> @dimabory Are you finished with your review?

We should resolve this -> https://github.com/ZEISS/precise-ui/pull/202#pullrequestreview-317497175

bajcmartinez:
> > @bajcmartinez could you sign the license/cla?
> > @dimabory Are you finished with your review?
> 
> We should resolve this -> [#202 (review)](https://github.com/ZEISS/precise-ui/pull/202#pullrequestreview-317497175)

I've pushed new changes with test coverage for those cases and the resolution

src/components/Highlight/Highlight.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  it('should render <Highlight> a single span if no match', () => {
+    const wrapper = shallow(<Highlight text="A" highlight="a" />);
+    expect(wrapper.children()).toHaveLength(1);
+  });
+
+  it('should render <Highlight> should escape \\', () => {
+    const wrapper = shallow(<Highlight text="Hello World!" highlight="\world" />);
+    expect(wrapper.children()).toHaveLength(1);
+  });
```
dimabory: Could you add snapshots for these cases?

bajcmartinez: Done

PR #208, opened 2019-12-16
feat(Highlight): Add highlight by matches

bajcmartinez:
### Prerequisites

### Contribution Type

### Description

Updates the Highlight component to allow the option to highlight text by matches, where a match is an array
of [start, end]


src/components/Highlight/index.tsx
```diff
@@ -10,7 +10,11 @@ export interface HighlightProps extends StandardProps {
   /**
    * String that needs to be highlighted
    */
-  highlight: string;
+  highlight?: string;
+  /**
+   * Array that defines the indices to highlight, each element is an array of [start, end]
```
dimabory: Could you also mention `highlight` is ignored in the case of `matches` provided?

src/components/Highlight/index.tsx
```diff
@@ -25,18 +29,41 @@ const Highlighted = styled.span`
...
  * Component will render a SPAN or series of SPAN with the content and highlights
  */
-export const Highlight: React.FC<HighlightProps> = ({ text, highlight, ignoreCase = true, theme }) => {
+export const Highlight: React.FC<HighlightProps> = ({ text, matches, highlight, ignoreCase = true, theme }) => {
+  if (!matches && !highlight) {
+    throw new Error('You must set either indices or highlight');
+  }
+
+  if (matches) {
+    let lastMatch = 0;
```
dimabory: Could you also validate `matches` to make sure it contains only numbers and the second one is more than first?

Some examples of odd behaviors:

image

image

image

image

image


```ts
const matches = "string"; // or {}, Date, function, etc.
<Highlight text="Hello World! Searching for World!" matches={matches} />

// TypeError: matches.map is not a function
```

PS:
It would be great to cover those cases by unit tests...

src/components/Highlight/index.tsx
```diff
@@ -25,18 +29,41 @@ const Highlighted = styled.span`
  * Highlight component is meant to be a simple component to display text with highlighted search.
  * Component will render a SPAN or series of SPAN with the content and highlights
  */
-export const Highlight: React.FC<HighlightProps> = ({ text, highlight, ignoreCase = true, theme }) => {
+export const Highlight: React.FC<HighlightProps> = ({ text, matches, highlight, ignoreCase = true, theme }) => {
+  if (!matches && !highlight) {
```
dimabory:
```suggestion
  if (!matches && undefined === highlight) {
```

You've broken this [invariant](https://github.com/ZEISS/precise-ui/pull/208/files#diff-45bf95c2583e477c07465a08c647e183R56-R58).

bajcmartinez: made the change, why exactly that way only for highlight?

bajcmartinez: nvm

src/components/Highlight/index.tsx
```diff
@@ -30,12 +31,29 @@ const Highlighted = styled.span`
...
-  if (!matches && !highlight) {
+  if (!matches && undefined === highlight) {
     throw new Error('You must set either indices or highlight');
   }
 
   if (matches) {
     let lastMatch = 0;
+    matches.forEach(match => {
+      if (
+        !Array.isArray(match) ||
```
dimabory: Could you extract this into the function `validateIndices(indices) { /*...*/ }`?
I would also throw for every case a respective message:

- [ ]  `(!Array.isArray(match) || match.length !== 2)` ->  `match must be an array of [start, end]`;
- [ ] `!Number.isInteger(match[0]) || !Number.isInteger(match[1])` -> `match must contain only positive integers`;
- [ ] `match[0] >= match[1]` -> `[start] must be lower than [end]`;
- [ ] `match[0] <= lastMatch` -> `the indices cannot overlap`.

dimabory: Still not fixed, right?
```tsx
const matches = "string"; // or {}, Date, function, etc.

<Highlight text="Hello World! Searching for World!" matches={matches} />

// TypeError: matches.map is not a function
```

bajcmartinez: I saw the comment about the messages coming, but I thought it would pass :p, pushing new changes now

src/components/Highlight/index.tsx
```diff
@@ -21,22 +26,76 @@ const Highlighted = styled.span`
...
 `;
 
+/**
+ * Validates that the matches provided are valid, if not it will throw an error
+ *
+ * @param matches
+ */
+function validateMatches(matches: Array<Array<number>>) {
+  let lastMatch = -1;
+
```
dimabory:
```suggestion

    if (!Array.isArray(matches)) {
      throw Error(`match must be an Array of [start, end], but provided 		${JSON.stringify(matches)}`);
    }

```

bajcmartinez:
> Could you update unit test snaps?

what should I do? tests are passing

dimabory:
> > Could you update unit test snaps?
> 
> what should I do? tests are passing

my apologies :)

dimabory:
LGTM 👍
Thank you, @bajcmartinez  🙂

Issue #209, opened 2020-01-03
Accordion header strips spaces when html tags are used


bajcmartinez:
### Prerequisites

- [X] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [X] Are you running the latest version?
- [X] Did you check the FAQs to see if that helps you?
- [X] Are you reporting to the correct repository?
- [X] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

### Versions

Latest - 0.8.1
Next - 0.8.3-pre.694

### Description

The accordion header strips spaces when having components with trailing spaces, provoking undesire behaviours

![image](https://user-images.githubusercontent.com/23080979/71722403-f0b22e00-2e28-11ea-9998-a11c013a7595.png)

### Steps to Reproduce

Use the following example:

```
const { Accordion, AccordionTab } = require('precise-ui');

<Accordion>
  <AccordionTab header={<><span>hello</span><span> world!</span></>}>
    {<><span>hello</span><span> world!</span></>}
  </AccordionTab>
</Accordion>
```

**Expected behavior:** You should read "hello world!"

**Actual behavior:** You read "helloworld!"

### Possible Solution

The issue seems to be happening because of the `display: flex` on the header component.

PR #211, opened 2020-01-09
fix(Accordion): header strips spaces when html tags are used

bajcmartinez:
fix(Accordion): header strips spaces when html tags are used

Use the following example:

```
<Accordion>
  <AccordionTab header={<><span>hello</span><span> world!</span></>}>
    {<><span>hello</span><span> world!</span></>}
  </AccordionTab>
</Accordion>
```

Expected behavior: You should read "hello world!"
Actual behavior: You read "helloworld!"

Closes: #209


dimabory:
Could you update CHANGELOG?

Issue #169, opened 2019-08-28
Dropdown menu options truncated for longer text.


MitaliDhamale:
**Description:**
The text is not visible completely in the dropdown menu, if the text to be selected is too long **IE11**

**Steps to reproduce:**
1. Try to select the option from the drop down which has larger text.

**Expected behavior:**
The option selected from the drop down should be visible(the complete text should be visible)

**Actual behavior:**
The option selected is not visible completely and seems distorted.

**Environment Details:**
IE11

**https://precise-ui.io/#/Components/DropdownSelect**

![DropdownOption](https://user-images.githubusercontent.com/52967194/63843758-36488100-c987-11e9-9e67-eef677add15a.png)

dimabory:
Reopening due to #212 https://github.com/ZEISS/precise-ui/pull/212

PR #212, opened 2020-01-10
Fixes: #201 DropdownMenu list has no width when expanded thus it's not possible to see the options on IE11

### LINKED ISSUE #201: DropdownMenu list has no width when expanded thus it's not possible to see the options on IE11
### Prerequisites

- [X] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [X] Are you running the latest version?
- [X] Did you check the FAQs to see if that helps you?
- [X] Are you reporting to the correct repository?
- [X] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

### Versions

0.8.2-pre.593 and up

### Description

When using DropdownMenu on IE11, after clicking on the menu to see the list of options the list appears with 0 width, which makes it impossible to visualize or select any option

### Steps to Reproduce

1. Render a DropdownMenu component
2. Open the page with IE11
3. Expand the menu
4. See the result

**Expected behavior:** [What you expected to happen]

To see the list and be able to select any of the options

**Actual behavior:** [What actually happened]

MicrosoftTeams-image

**Environment details:** [OS, Browser, ...]

IE11

### Possible Solution
The issue was introduced in commit e47590ae14162eecaf0361f6d7ea1215d7843470 particularly in `ListItemContent` change from

```
const ListItemContent = styled.div`
  flex-grow: 1;
`;
```

to

```
const ListItemContent = styled.div`
 flex-grow: 1;
 flex: 1 0 0px;
`;r
```

The change adding flex-basis to 0px broke the IE11 compatibility.

gerkirill:
### Prerequisites

### Contribution Type

### Description

Fixes #201 - DropdownMenu and SelectButton when expanded - used to have an issue in IE11: the options list has had zero width.


cz-white:
@dimabory  I am going to disagree with your comment here, the dropdown menu is a separate issue that (if it hasn't already) should be raised. 

#169 refers to the dropdownSelect component, not the dropdownMenu component which with the fix from @gerkirill, still works as expected.  

the issue you attached a screenshot of has to do with the dropdownMenu component specifically having no-wrap set to it, and with a max-width of 300px. I am assuming this is intentional to limit the size of such a menu, and the items in it. If we want the text to wrap here that component will need a larger refactor, as its width appears to be programmatically set, which leads to undesirable wrapping behavior. 

TLDR: I think @gerkirill s change fixes the issue at hand, does not impact #169 and a new issue may need to be raised for the dropdownMenu if it is decided to actually be an issue.

guiherzog:
I also support @cz-white's opinion. Therefore, we will be merging this fix. Feel free to create an issue for the other problem.

Issue #225, opened 2020-01-20
UploaderProgressDetails opens behind first Modal in IE


rad3k1:
### Description
If FileUploader with FileUploaderDetails is called from antoher Modal (1), 
the UploaderProgressDetails (2) opens behind that modal in Internet Explorer (it works fine in Chrome and FF).

![upload_details](https://user-images.githubusercontent.com/53941899/72711562-8c51d700-3b69-11ea-9d22-27c4b572cb4b.PNG)

**Expected behavior:**
UploaderProgressDetails should appear in front of the first modal in IE.

**Actual behavior:**
UploaderProgressDetails opens behind the first modal in IE.

**Environment details:**
Browser: IE11

### Possible Solution
The problem is "z-index: 10000", which is set in "Blocker" component (called by Modal) with following styled component:
```
const FixedContainer = styled.div`
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;
  position: fixed;
  z-index: 10000;
  overflow-x: hidden;
  overflow-y: auto;
`;
```

One solution is to override this property in "FileUploaderDetails" component in existing styled component "StyledDetailsHost":
```
const StyledDetailsHost = styled.div`
  & > div:first-child {
    z-index: 10001;
  }
`;
```

PR #226, opened 2020-01-20
Bugfix/225 fix opening upload progress details in ie

rad3k1:
### Prerequisites

### Contribution Type

### Description

Bugfix for [issue 225](https://github.com/ZEISS/precise-ui/issues/225).

The problem is "z-index: 10000", which is set in "Blocker" component (called by Modal) with following styled component:
```
const FixedContainer = styled.div`
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;
  position: fixed;
  z-index: 10000;
  overflow-x: hidden;
  overflow-y: auto;
`;
```

This property will be overridden in "FileUploaderDetails" with higher value (10001).


src/components/FileUploaderDetails/index.tsx
```diff
@@ -64,7 +64,11 @@ const StyledUploaderHost = styled.div`
 `;
 
 const StyledDetailsHost = styled.div`
-  z-index: 10001;
+  // We override the 'z-index' of the styled component 'FixedContainer' of 'Blocker' (which is called by 'Modal').
+  // This fixes the problem of showing 'UploadProgressDetails' in front of 'parent' modal in IE.
```
guiherzog: This explanation is confusing, isn't it supposed to say, this fixes the problem of showing ... behind modal in IE? Also, file UploadProgressDetails is not really a child of the modal. Probably a sibling. 

The fix seems hacky, but that's ok for now, but the explanation is a big confusing.

Issue #228, opened 2020-01-21
onFocus gets triggered on AutoComplete on every key stroke.


guiherzog:
### Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository?
- [x] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

### Versions

Precise-UI: 0.8.1

### Description

Every time a key is pressed on the Autocomplete component, the `onFocus` method gets called. 

### Steps to Reproduce

1. Go to Kitchen Sink
2. Log the onFocus method.
3. Type a few keys in the input.
4. The log will be printed the same amount of keys pressed

**Expected behavior:** onFocus gets called only when focusing

**Actual behavior:** onFocus gets called for every keystroke

**Environment details:** Chrome 79, MacOS

benbouba:
This repository is being archived, and all issues have been closed as part of the process.

PR #231, opened 2020-01-21
Removing refocus on every `onChange` Event in Autocomplete

Saladinek:
### Prerequisites

### Contribution Type

### Description

Removing refocus on every onChange Event as it breaks the very principle of onFocus functionality. 
Also removing corresponding test case. This PR corresponds to issue [#228 ](https://github.com/ZEISS/precise-ui/issues/228)

For now adding "Breaking Change" option, as it changes the current 'intended' functionality.


src/components/Autocomplete/index.tsx
```diff
@@ -181,24 +181,7 @@ class AutocompleteInt<T> extends React.Component<AutocompleteProps<T> & FormCont
...
-        If we call this.hide() again with a timeout, it works also for IE11.
-      */
-      setTimeout(() => {
-        if (this.state.open) {
-          this.hide();
-        }
-      });
-    };
-
-    suggestionSelected ? hide() : this.show();
```
dimabory: unfortunately, it's not a solution. it will bring another issue https://github.com/ZEISS/precise-ui/issues/123

dimabory: see also https://github.com/ZEISS/precise-ui/pull/184

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Precise UI Changelog
 
+## 1.0.1
+
+- Fix Accordion not responding when defaultSelectedIndex is set
+- Remove refocus of component(Autocomplete) on every value change, remove corresponding test
```
dimabory: this went to `0.8.3`(`1.0.0`).

src/components/Autocomplete/index.tsx
```diff
@@ -249,14 +232,9 @@ class AutocompleteInt<T> extends React.Component<AutocompleteProps<T> & FormCont
...
-
-    this.setState(
-      () => ({
-        open: true,
-      }),
-      onFocus,
-    );
+    this.setState(() => ({
+      open: true,
+    }));
```
dimabory: no need to use a factory. just `this.setState({ open: true })` looks better ;)

Saladinek: Fixed.

CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
 # Precise UI Changelog
 
 ## 1.0.1
+
 - Fix opening `UploaderProgressDetails` behind modal in IE (#225)
+- Remove refocus of component(Autocomplete) on every value change, remove corresponding test
```
dimabory: I just updated origin. It's no more `1.0.1`. Can you update the version and changelog?

dimabory: Also seems like it is kind of **Breaking Change**, but I wouldn't say it complete change the functionality, it rather fixes the issue, so probably we can bump up the minor version.

dimabory: > Also seems like it is kind of Breaking Change, ...

> ..., so probably we can bump up the minor version

@Saladinek Could you update a minor instead of a patch?

PR #277, opened 2020-04-15
Add custom dialog for Prompt component

magneticz:
Current implementation of Prompt component uses only browser's system dialog. This update brings possibility to use Modal Dialog from precise-ui as well as gives more flexibility with `usePrompt` hook, if modal needs to be further customized

### Prerequisites

### Contribution Type

### Description

[Place a meaningful description here.]

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "precise-ui",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "description": "Precise UI React component library.",
   "main": "dist/es6",
   "typings": "dist/es6",
--- a/src/components/Form/Example.md
+++ b/src/components/Form/Example.md
@@ -137,8 +137,8 @@ The `Form` with validation rules
 ```jsx
 const { Form, Button, TextField } = require('precise-ui');
 
-<Form 
-  onSubmit={e => alert(JSON.stringify(e))} 
+<Form
+  onSubmit={e => alert(JSON.stringify(e))}
   validationRules={{
     first: (value) => value && value.length > 10 ? 'Should be less than 10' : undefined,
     last: () => 'Always some error',
@@ -160,3 +160,33 @@ const { Form, Button, TextField } = require('precise-ui');
   </div>
 </Form>
 ```
+
+**Using Prompt with Form**
+
+When form has changes, it is possible to make a prompt, if user wants to leave the page without submitting the form. `Form` component has built-in `Prompt` component.
+
+Simple way to use it is to provide `prompt` prop. It will show browser's system dialog.
+
+```jsx  { "props": { "data-skip": true } }
+const { Form } = require('precise-ui');
+
+<Form prompt="Form has unsaved changes, do you want to leave?">
+{/* other components*/}
+</Form>
+```
+
+It is also possible to provide `Prompt` component, that will show `precise-ui` component instead of system dialog.
+
+```jsx  { "props": { "data-skip": true } }
+const { Form, Prompt } = require('precise-ui');
+
+<Form prompt={(changed: boolean) => {
+  return (
+    <Prompt message="Your form has unsaved messages, are you sure you want to leave the page" when={changed} modalOptions={{ title: 'Do you want to leave' }} />
+  );
+}>
+{/* other components*/}
+</Form>
+```
+
+For more details refer to `Prompt` component examples
--- a/src/components/Form/index.tsx
+++ b/src/components/Form/index.tsx
@@ -42,9 +42,9 @@ export interface FormValidationError {
 export interface FormProps<FormValues> extends StandardProps {
   /**
    * Shows the given message if the user wants to navigate
-   * with changes being made.
+   * with changes being made or renders custom component with message if provided.
    */
-  prompt?: string;
+  prompt?: ((changed: boolean) => React.ReactChild) | string;
   /**
    * The value of the form to be used in controlled mode.
    */
@@ -320,7 +320,7 @@ export class Form<Values extends FormValuesData> extends React.Component<FormPro
     const { changed } = this.state;
     return (
       <StyledForm {...rest} onSubmit={this.submit}>
-        {prompt && <Prompt when={changed} message={prompt} />}
+        {prompt && (typeof prompt === 'function' ? prompt(changed) : <Prompt when={changed} message={prompt} />)}
         <FormContext.Provider value={this.ctx}>{children}</FormContext.Provider>
       </StyledForm>
     );
--- a/src/components/Prompt/Example.md
+++ b/src/components/Prompt/Example.md
@@ -1,0 +1,83 @@
+`Prompt` component allows to show a dialog window if user is navigating away from a page. Usual use case for this is when user is about to leave the page with form that has changes without submitting it.
+
+**Note:** For `Prompt` component to work, it needs to have `router` from react-router through context provider. If react-router context is not setup, `PromptBasic`, `PromptModal` or `usePrompt` can be used instead. Look at **Advanced** section.
+
+**Important:** Message passed to the `Prompt` component will be shown during page transitions within the app. If the transition is happening outside of the app (moving to different url host, reloading the page), then browser's system dialog, with system message will be shown. This message cannot be customized.
+
+**Default Mode**
+
+By default, `Prompt` component will display browser's dialog window with custom message.
+
+```jsx { "props": { "data-skip": true } }
+const { Prompt } = require('precise-ui');
+
+<Prompt message="Hey, are you sure you want to leave the page" />
+```
+
+**Custom Modal Mode**
+
+`Prompt` component can also display custom component when navigation between pages inside application.
+
+
+```jsx { "props": { "data-skip": true } }
+const { Prompt } = require('precise-ui');
+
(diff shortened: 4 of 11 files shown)


src/components/Form/index.tsx
```diff
@@ -42,9 +42,9 @@ export interface FormValidationError {
 export interface FormProps<FormValues> extends StandardProps {
   /**
    * Shows the given message if the user wants to navigate
-   * with changes being made.
+   * with changes being made or renders custom component with message if provided.
    */
-  prompt?: string;
+  prompt?: (changed: boolean) => React.ReactChild | string;
```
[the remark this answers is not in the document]
magneticz: it doesn't have to. It compiles without any problems, and recognizes types properly.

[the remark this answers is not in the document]
dimabory: [t.ly/yZEXD](url)
```ts
type Foo = (changed: boolean) => {} | string;
type Bar = ((changed: boolean) => {}) | string;

const foo: Foo = ''; // Type '""' is not assignable to type 'Foo'.(2322)
const bar: Bar = '';
```

I guess Ivan's point is right. It will compile well for sure, but you won't be able to pass a string as a value of `prompt` using TS.

dimabory, line 47: The example with a simple prompt is no longer working for a Form component.
```ts
<Form prompt="Are you sure?">
 <TextField/>
 <Button>Ok</Button>
</Form>
```

I didn't dig into that to check the reason. Could you verify it?

PS: If it's related to necessary to include `react-router` to make it work within a form then it is a breaking change.

magneticz: good catch. The general page reload, or leaving the app stopped working. The reason, I forgot that it is supported regardless, if there is react-router or not. I will update my PR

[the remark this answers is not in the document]
magneticz: I see what you mean now, will fix it. Thanks @IKolosynskyi and @dimabory

src/components/Prompt/Example.md
```diff
@@ -0,0 +1,74 @@
+`Prompt` component allows to show a dialog window if user is navigating away from a page. Usual use case for this is when user is about to leave the page with form that has changes without submitting it.
```
dimabory, line 1: I think the main example should be with a Form component. This is the most used example I suppose. I don't see such one.
Another one would be related to FileUploaderDetails.
The current solution still forces developers to write a lot of custom code to implement a prompt.

The best scenario could be using a dedicated HOC...
```tsx
// import { withPrompt } from 'precise-io';

// 1. usage with Form
const FormWithPrompt = withPrompt(Form);
<FormWithPrompt onSubmit={console.log}/>

// 2. usage with whatever else
const MyComponent = () => <div>My Component</div>
const MyComponentPrompt = withPrompt(MyComponent);
<MyComponentPrompt />
```

The main question here is how to deal with nested `withPrompt` components. For instance
```ts
const Prompt1Component = withPrompt(({children}) => <div>{{children}}</div>);
const Prompt2Component = withPrompt(() => <div>Some text</div>);
<App>
 <Prompt1Component>
  <Prompt2Component/>
 </Prompt1Component>
</App>
```

magneticz: Example can be added.

Regarding the a lot of custom code, I will have to disagree. The easiest is to just drop the `Prompt` component in your page and it will just work, as long as you have react-router context set higher above. If you don't, you can use `PromptBasic` or `PromptModal` (depending if you want browser's default prompt window or modal) and pass in text and history object. 
```jsx
// will always prompt, when you want to leave the page. 
<Prompt message="Hey, are you sure you want to leave the page" />


// this is the equivalent for 
//  const FormWithPrompt = withPrompt(Form);
// <FormWithPrompt onSubmit={console.log}/>
<Form prompt="Hey, are you sure you want to leave the page"">


// this is with precise-ui Modal
<Prompt message="Hey, are you sure you want to leave the page" modalOptions={{title: "Leaving page..."}}/>
```
For examples above, you just need to drop those components and it will work, but you also have more control (what is the text, when to show etc) in more convenient way(component rather than HOC)

And if user needs to have more customized thing, where he doesn't need modal or prompt, but button appearing on page for example, then he can just use `usePrompt` and have fully customized version.

To achieve similar things, HOC will not save a lot of custom code(maybe none), but also make it harder to control, when to enable/disable the Prompt component

regarding the question about nested components, this component has to be used with caution, since it can easily get complicated, like in scenario you wrote. 

I can give another scenario, that this component becomes problematic.
Image, you have a page, with two tabs. one tab contains list of items, second tab contains the Form. You set the form to have a prompt, so if user leaves edited form, it will ask if user really wants that. And most importantly, when user switches between tabs, url changes.

In this case, when user changes the form, and navigates to second tab, prompt will be asked, if user wants to leave, he presses ok, and navigates to second page. But, components in the first tab with form are not unmounted, the form is changed and not submitted and etc. So, when user decides to navigate again, he will be asked again, if he wants to navigate away. 

So, it is important to use the `when` prop wisely, so that prompts do not overlap, or block user's navigation to other pages

src/components/Form/index.tsx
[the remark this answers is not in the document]
magneticz: updated

src/components/Form/Example.md
```diff
@@ -160,3 +160,33 @@ const { Form, Button, TextField } = require('precise-ui');
...
 </Form>
 ```
+
+**Using Prompt with Form**
+
+When form has changes, it is possible to make a prompt, if user wants to leave the page without submitting the form. `Form` component has built-in `Prompt` component.
+
+Simple way to use it is to provide `prompt` prop. It will show browser's system dialog.
+
+```jsx
```
dimabory, line 170: There is no sense to make a snapshot of this example.

```suggestion
```jsx { "props": { "data-skip": true } }
```

src/components/Form/Example.md
```diff
@@ -160,3 +160,33 @@ const { Form, Button, TextField } = require('precise-ui');
...
+const { Form } = require('precise-ui');
+
+<Form prompt="Form has unsaved changes, do you want to leave?">
+{/* other components*/}
+</Form>
+```
+
+It is also possible to provide `Prompt` component, that will show `precise-ui` component instead of system dialog.
+
+```jsx
```
dimabory, line 180:
```suggestion
```jsx { "props": { "data-skip": true } }
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "precise-ui",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
dimabory, line 3: I would rather change a minor version as you've added new functionality.

src/components/Prompt/Prompt.types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export interface PromptHistoryBlockCallback {
+  (cb: (location: Location) => any): PromptCallback;
+}
+
+export interface PromptCallback {
+  (): boolean | string;
+}
+
+export interface PropDefaultModalOptions {
```
dimabory, line 59: 1. do you need to export it?
2. the naming a bit confusing me. does `BasePromptModalOptions` look better for you?

magneticz: there is a type, it should be `PromptDefaultModalOptions`
does `PromptDefaultModalOptions` make more sense? these are options for default modal, provided by the Prompt component

dimabory: is it necessary to export?

src/components/Prompt/Example.md
dimabory: >  then he can just use usePrompt and have fully customized version.


could you also add example on how to use `usePrompt`?

src/components/Prompt/Prompt.types.ts
magneticz: not at the moment, but it might be useful when using Prompt with modal, to pass modal options

src/components/Prompt/Example.md
```diff
@@ -0,0 +1,83 @@
...
+    <br/>
+    <ActionLink onClick={simulateNavigation}>Simulate navigation to differnt page</ActionLink>
+    <PromptModal when={enable} history={history} message="Your Form contains unsaved data" modalOptions={{title: "Are you sure you want to leave", confirmText: "Leave", cancelText: "Stay", onConfirm: ()=>{console.log("user is navigating to different page")}, onCancel: ()=>{console.log("user decided to stay in the page")}}} /></>);
+  }
+}
+
+<PromptExample/>
+```
+
+To create custom Modal (or any other functionality that ask before navigating to different page), `usePrompt` hook can be used. For details on how to use look at implementation of `PromptModal` component (`PromptModal.tsx`)
```
magneticz, line 83: @dimabory, regarding the example for `usePrompt`, i think it will be redundant as `PromptModal` is implemented with `usePrompt` and is a very good showcase. Referring to that component is enough i think

src/components/Prompt/Example.md
magneticz: replied as another comment below

src/components/Prompt/Example.md
dimabory: okey dokey ;)

Issue #332, opened 2021-07-13
Grouped elements in Table aren't sorted properly


dremartins:
### Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository?
- [x] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

### Versions

PreciseUI 1.5.1
You can get this information from executing `npm version`.

### Description

Sorting isn't always working when combining groups and key sort on Table or AccordionTable components

### Steps to Reproduce

1. Use following code on kitchensink:
```
const { AccordionTable } = require('precise-ui');
const style = { width: '200px', margin: '0 auto', textAlign: 'center' };

function getContent({index, data}) {
  return (
    <div style={style}>
      Details for row #
      <b>{index}</b>
    </div>
  );
}

<AccordionTable
  groupBy="tag"
  columns={{tag:{header: 'tag'}, value: {header: 'value', sortable:  true}, team: {header: 'team'}}}
  detailsRenderer={getContent}
  data={[
    {tag: 'A', value: 'Alpha1', team: 'Alpha team'},
    {tag: 'A', value: 'Alpha3', team: 'Alpha team'},
    {tag: 'B', value: 'Bravo2', team: 'Bravo team'},
    {tag: 'A', value: 'Alpha2', team: 'Alpha team'},
    {tag: 'B', value: 'Bravo3', team: 'Bravo team'},
    {tag: 'C', value: 'Charlie1', team: 'Charlie team'},
    {tag: 'E', value: 'Echo1', team: 'Echo team'},
    {tag: 'A', value: 'Alpha4', team: 'Alpha team'},
    {tag: 'B', value: 'Bravo1', team: 'Bravo team'},
    {tag: 'E', value: 'Echo2', team: 'Echo team'},
  ]} />
```
2. Expand the groups
3. Sort by value descending
4. Feel free to replace the AccordionTable with simple Table. Result is the same

**Expected behavior:** 
All groups should be maintained and elements sorted by value.

**Actual behavior:** 
Not all elements are properly sorted. Group B elements in this example are wrongly sorted.

**Environment details:** any

### Possible Solution

Fix sorting by multiple keys (sort key and group key)

PR #333, opened 2021-07-14
Bugfix/332 table sorting with groups

dremartins:
### Prerequisites

### Contribution Type

### Description

Fix for #332. When grouping and sorting are combined running sort twice (first for sorting, second for grouping) would not guarantee the stable order of the elements, casing grouped elements to possibly become unsorted.
A unit test was expanded to break the previous version, and both an example on `AccordionTable` and `Table` were updated to show this.


src/utils/sort.ts
```diff
@@ -128,12 +148,12 @@ export function sortObjectList<T extends {}>(
...
-    sorter(result, items, sortBy, order === 'descending');
+    sorter(result, items, sortBy, groupBy, order === 'descending');
+  } else if (groupBy) {
+    sorter(result, items, groupBy, undefined, order === 'descending');
   }
 
-  if (groupBy && groupBy !== sortBy) {
-    sorter(result, items, groupBy);
-  }
+  console.log('RESULT: ', result);
```
boromo, line 156: I guess we don't need this log?

dremartins: Damn... there is always one hiding somewhere...

boromo (approved):

PR #348, opened 2021-09-14
Fix/566349 implement Flyout component with popper

simenko:
### Prerequisites

### Contribution Type

### Description

The Flyout component was re-implemented using [Popper.js](https://popper.js.org/). The original implementation contained several hundred lines of code responsible for Flyout positioning. The code was quite complex, had bugs in it, there were no unit tests, so it was really hard to maintain. 

Popper.js is a mature positioning engine with almost 4 million weekly downloads on npmjs.com and 18K stars on Github. It allowed to completely get rid of custom positioning code. The component source code was reduced from around 600 lines to 300 (~100 of them is required only for compatibility with current implementation), bundle size was reduced by 4 kilobytes (because we already have Popper.js as a dependency of react-datepicker).

The public API of the Flyout component remained unchanged, but slight changes in the Flyout position and behaviour are possible. We could think about changing public API to get rid of some compatibility overhead and be able to use all the features offered by Popper.

Th flyoutArrowSize constant was moved from the component code to the theme.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Precise UI Changelog
 
+## 2.1.0
```
FlorianRappl, line 3: Either merge with 2.0.1 to 2.1.0 or append it to 2.0.1 (I think it should be merged).

Note that 2.0.1 was not released yet.

package.json
```diff
@@ -43,10 +43,12 @@
     }
   },
   "dependencies": {
-    "date-fns": "^2.3.0",
+    "@popperjs/core": "^2.10.1",
+    "date-fns": "^2.23.0",
     "memoize-one": "^5.1.0",
     "react-datepicker": "^4.2.1",
     "react-onclickoutside": "^6.7.1",
+    "react-popper": "^2.2.5",
```
FlorianRappl, line 51: Ok so we gain 4 KB (nice), but we loose

- `react-popper` (5.4 KB)
- `@popperjs/core` (20 KB)

A total of 25 KB - so 21 KB more.

See https://bundlephobia.com/package/@popperjs/core@2.10.1 and https://bundlephobia.com/package/react-popper@2.2.5.

Am I missing something?

src/utils/flyoutCompatibilityHelpers.tsx
```diff
@@ -0,0 +1,57 @@
...
+  if (!transform || !popperAttributes['data-popper-placement']) {
+    return arrowStyles;
+  }
+
+  const primaryPlacement = (popperAttributes['data-popper-placement'] as Placement).split('-')[0];
+
+  const placementDependentStyles: React.CSSProperties = {
+    top: {
+      bottom: 0,
+      transform: transform + ` rotate(${0}deg)`,
```
FlorianRappl, line 40: If you do start a template string, why not make it a full template? --> `${transform} rotate(0deg)` (why the extra `${}` if you insert a constant anyway?

FlorianRappl (commented):
Fine in general - but I am not sure what those bugs are (can you name some) and the additional weight + dep maintenance is truly required if its only about the flyout. Could we use this at other positions?

Also assuming the mentioned relation between popper and datepicker is true - why do we need to have the additional deps anyway (right now that could only result in multiple versions being resolved --> see dep maintenance)?

package.json
FlorianRappl: Alright but then I'd recommend to use `^2.9.2` for the `@popperjs/core` version range, to have the same base quantifier (resolved anyway to 2.10.1 currently).

simenko: Good catch, done!

simenko:
> Fine in general - but I am not sure what those bugs are (can you name some) and the additional weight + dep maintenance is truly required if its only about the flyout. Could we use this at other positions?

The existing Flyout component was not working properly in the narrow viewport (e.g. on mobile browsers) - large portions of it were not visible, and it could not be fixed by tweaking the position prop - depending on current language setting, viewport width, and zoom the target component could be positioned anywhere on the screen. If we could make it look OK either on the right or on the left side, but not anywhere. This PR not only fixes that issue but also removes a whole lot of complexity from our codebase and reduces bundle size a bit.

The Flyout component is used not just directly but also by Tooltip, DropdownMenu, and OverflowMenu components.
 
> Also assuming the mentioned relation between popper and datepicker is true - why do we need to have the additional deps anyway (right now that could only result in multiple versions being resolved --> see dep maintenance)?

While I could use the Popper deps without adding them to the dependencies list explicitly, I believe it is a bad practice. If we update react-datepicker or replace it with some other date picker package, we can unexpectedly break the Flyout.

FlorianRappl:
Thanks for the explanation.

> While I could use the Popper deps without adding them to the dependencies list explicitly, I believe it is a bad practice. If we update react-datepicker or replace it with some other date picker package, we can unexpectedly break the Flyout.

I think the intention has been misunderstood. Again, this was not about hiding the dependency, but about the additional maintenance cost. Every dependency adds a hidden cost and needs to be accommodated. We should make sure these new direct dependencies don't go out of alignment with the already present indirect dependencies.

src/utils/flyoutCompatibilityHelpers.tsx
```diff
@@ -0,0 +1,57 @@
...
+  if (!transform || !popperAttributes['data-popper-placement']) {
+    return arrowStyles;
+  }
+
+  const primaryPlacement = (popperAttributes['data-popper-placement'] as Placement).split('-')[0];
+
+  const placementDependentStyles: React.CSSProperties = {
+    top: {
+      bottom: 0,
+      transform: `${transform} rotate(${0}deg)`,
```
FlorianRappl, line 40: Can we optimize these templates?

simenko:
> We should make sure these new direct dependencies don't go out of alignment with the already present indirect dependencies.

Take a look at the last 2 commits. I tried to automate dependency versions check. At first, I hoped to use `npm find-dupes`, but it scans the local node_modules directory and its output is inconsistent. So I have implemented a simple script that checks versions of production dependencies all over the deps tree and included it in the husky pre-push hook. 

I have already found another version of react-onclickoutside with the help of this script and updated it. 

What do you think?

package.json
```diff
@@ -39,14 +39,16 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "npm run test & npm run build"
+      "pre-push": "node tools/deps-duplication-detector.js && (npm run test & npm run build)"
```
FlorianRappl, line 42: We maybe should be the dup checker into the scripts an just have `npm run check` or so here.

FlorianRappl (commented):
This looks great - thanks :beers:!

PR #355, opened 2021-12-20
Bug/600319

t-tkachenko:
### Prerequisites

- [x ] I have read the **CONTRIBUTING** document
- [x ] My code follows the code style of this project

### Contribution Type

- [x ] Bug fix (non-breaking change which fixes an issue, please reference the issue id)
- [ x] I have updated the documentation accordingly
- [ x] All new and existing tests passed

### Description

_Carousel_ bug fixed: (#600319: see details [here](https://dev.azure.com/ZEISSgroup/DCC/_workitems/edit/600319)) iBase Announcements | Navigation dots shown on slider even for 1 page

### DIFF
--- a/src/components/Carousel/index.tsx
+++ b/src/components/Carousel/index.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import styled, { css } from '../../utils/styled';
-import { KeyCodes } from '../../utils/keyCodes';
-import { remCalc } from '../../utils/remCalc';
+import { KeyCodes, remCalc } from '../../utils';
 import { InteractiveSurface, InteractiveSurfaceChangeEvent, InteractiveSurfaceProps } from '../InteractiveSurface';
 import { Icon } from '../Icon';
 import { StandardProps } from '../../common';
@@ -452,6 +451,7 @@ export class Carousel extends React.PureComponent<CarouselProps, CarouselState> 
 
     const disableLeft = !infinite && selectedIndex < 1;
     const disableRight = !infinite && selectedIndex > childrenCount - 2;
+    const enableBullets = bullets && bullets.length > 1;
 
     return (
       <RootContainer {...props} onKeyDown={this.handleKeyDown} tabIndex={0}>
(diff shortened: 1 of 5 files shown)


src/components/Carousel/index.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import styled, { css } from '../../utils/styled';
-import { KeyCodes } from '../../utils/keyCodes';
-import { remCalc } from '../../utils/remCalc';
+import { KeyCodes, remCalc } from '../../utils';
```
FlorianRappl, line 3: Not 100% sure about this (there might have been a reason for having explicit module imports instead of taking the `index`, maybe one of the utils refers to components which would make it cyclic?). Have you checked that this works?

If yes, then fine with me.

t-tkachenko: Yes I'm 100% sure it works cause I've checked it on local machine (put my case into runnable docs) and code has passed all the tests.

teberl:
I think it would be preferable if the component exposes a property, like `pagination` or `bullets` which is of type `boolean` that way the parent could handle the visibility itself.
That approach would make it more flexible to use and also enable the option to completly disable the bullet points if desired.

t-tkachenko:
> Thanks @t-tkachenko for the contribution. IMHO it would be preferable if the component exposes a property, like `pagination` or `bullets` which is of type `boolean` that way the parent could handle the visibility itself. Default would be `true`, to not break existing behaviour. That approach would make it more flexible to use and also enable the option to completly disable the bullet points if desired.
> 
> Maybe this is also not a bug, i feel it is more like a feature request for the component?

In the current state it is definitely a bug. If Carousel has only one Page it means that it's no places (and points) to navigate with bullets, arrows, etc. So navigation facilities should be hidden.
Your proposal is a feature allowing parent component to hide Bullets on demand no matter the Pages count.
However it's not relied to this bug.

benbouba:
This repository is being archived, and all PRs have been closed as part of the process.

PR #371, opened 2022-03-01
Fix: move @types/react-datepicker from devDependencies

panichevoleg:
### Prerequisites

- [ ] I have read the **CONTRIBUTING** document
- [ ] My code follows the code style of this project

### Contribution Type

### Description

If `DateFeild` is used in zeisslet, it can't be built when `@types/react-datepicker` is in `devDependencies`. I have moved it to `dependencies`

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.9.2",
+    "@types/react-datepicker": "^4.1.5",
     "date-fns": "^2.23.0",
     "memoize-one": "^5.1.0",
     "react-datepicker": "^4.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.9.2",
+    "@types/react-datepicker": "^4.1.5",
```
FlorianRappl, line 48: Any reason for moving it here? Usually, typings are dev dependencies as they are not required at runtime. Maybe have them as `peerDependencies`? Then they would still be automatically resolved if someone does not have them, but they are not indicated as required per se.

FlorianRappl:
> If DateFeild is used in zeisslet, it can't be built when @types/react-datepicker is in devDependencies. I have moved it to dependencies

Isn't that rather the job of the Zeisslet? If a Zeisslet is missing types then it should add these types.

package.json
panichevoleg: You are right. But I need it not at runtime, I need it in development (when use `DateField` component). When it's in `dependencies` - it's not a part of precise-ui installed into zeisslet deps.

FlorianRappl: Yes, but that is not a problem of precise-ui. If every library would need to care about the dev setup of any consumer app nothing would ever ship.

The solution is that the Zeisslet needs to install the right typings. After all, it seems to care about these typings (for some reason). Otherwise, I'd recommend not to expose third-party typings.

PR #372, opened 2022-03-16
WCAG: PreciseUI table has wrong header

panichevoleg:
### Prerequisites

### Contribution Type

### Description

Accoring to WCAG 2 validator, there is `Empty table header` error.

**What does the Empty Table Header error mean?**

An empty table header error means that one of the table headers on your post or page does not contain any text. This means that the `<th>` element is present but looks like this `<th></th>` with nothing between the opening and closing tags.

**How to Fix It**

If the table cell is a header, provide text within the cell that describes the column or row. **If the cell is not a header or must remain empty (such as the top-left cell in a data table), make the cell a `<td>` rather than a `<th>`**.

So, if `<th></th>` is empty, I render it as `<td></td>` (styling kept the same).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Precise UI Changelog
 
+## 2.1.10
+
+- Fix WCAG error: Empty table header
+
 ## 2.1.9
 
 - Fixed `arrowToggle` on `AccordionTable`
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
LGTM - great job!
:beers:

PR #375, opened 2022-04-21
Fix the issue `The component Styled(styled.span) has been created dynamically

panichevoleg:
### Prerequisites

### Contribution Type

### Description

`Skeleton` component produces the error:

`The component Styled(styled.span) with the id of "sc-fUOEQR" has been created dynamically.
You may see this warning because you've called styled inside another component.
To resolve this only create new StyledComponents outside of any render method and function component.`

The solution is to move creation of `StyledSpan` component out of `render()` method.

### DIFF
--- a/src/components/Skeleton/index.tsx
+++ b/src/components/Skeleton/index.tsx
@@ -66,8 +66,8 @@ const formatValue = (value: string | number) => {
 };
 
 const StyledSpan = styled(Span)<SkeletonProps>`
-  ${({ count = 1, duration = 1.2, width = '100%', height = '100%', isCircle, isText, isPulsing = true }) => css`
-    animation: ${shine} ${duration}s infinite linear ${isPulsing ? 'running' : 'running'};
+  ${({ count, duration = 1.2, width = '100%', height = '100%', isCircle, isText, isPulsing = true }) => css`
+    animation: ${shine} ${duration}s infinite linear ${isPulsing ? 'running' : 'paused'};
     width: ${count > 1 && isText ? `${randomInt(80, 100)}%` : formatValue(width)};
     height: ${formatValue(height)};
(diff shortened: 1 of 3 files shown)


src/components/Skeleton/index.tsx
```diff
@@ -61,30 +61,27 @@ const Span = styled.span`
 
 const randomInt = (min: number, max: number) => Math.floor(Math.random() * (max - min + 1)) + min;
 
+const formatValue = (value: string | number) => {
+  return typeof value === 'number' ? `${value}px` : value;
+};
+
+const StyledSpan = styled(Span)<SkeletonProps>`
+  ${({ count = 1, duration = 1.2, width = '100%', height = '100%', isCircle, isText, isPulsing = true }) => css`
```
klaidigorishti, line 69: Do you need to add a default `count` here as well? There is already a default coming from `props` in line #81.

src/components/Skeleton/index.tsx
```diff
@@ -61,30 +61,27 @@ const Span = styled.span`
...
+    animation: ${shine} ${duration}s infinite linear ${isPulsing ? 'running' : 'running'};
```
klaidigorishti, line 70: `isPulsing ? 'running' : 'paused'`

src/components/Skeleton/index.tsx
panichevoleg: It's not coming from there, I use it there just locally, as max value of loop. But ok, I can pass that value to `StyledSpan` and overwrite value from props.

src/components/Skeleton/index.tsx
panichevoleg: Right :)

klaidigorishti (approved):
LGTM! Thanks

PR #380, opened 2022-06-22
Improve validation of Form

panichevoleg:
### Prerequisites

### Contribution Type

### Description

There are cases when conditional validation of form fields is needed. E.g.,:
- only one of fields need to be filled
- field need to be filled when checkbox is set
- etc
Currently such kind of validation is not supported.

Current PR solves this issue in the following way:
- validator function gets not only value to e validated, but also data of the whole form
- new optional prop `validateWith` is introduced for all form controls - there can be passed list of fields that need to be re-validated when form value is changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Precise UI Changelog
 
+## 2.1.14
+
+- Improve validation of `Form`
+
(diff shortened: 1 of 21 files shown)


FlorianRappl (approved):
LGTM - thanks!

PR #374, opened 2022-03-24
#373 add page size event to pagination component

peterugah:
### Prerequisites

Please make sure you can check the following two boxes:

- [x] I have read the **CONTRIBUTING** document
- [x] My code follows the code style of this project

### Contribution Type

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue, please reference the issue id)
- [x] New feature (non-breaking change which adds functionality, make sure to open an associated issue first)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

### Description

In this pull request, I added the ability for the Pagination component to pass the on size changed event as a prop that can be listened for. This is required to correctly compute pagination when pagination is handled in a controlled manner.

PR #381, opened 2022-07-12
feat: fix react error from onSizeChange in the Pagination component

feedm3:
### Prerequisites

### Contribution Type

### Description

Currently, the `onSizeChange` handler gets passed down with prop spreading to a different component. This leads to an error in the console, as the prop gets attached to a basic HTML element.
grafik

This fix removes the handler from the standard props before they are passed down.

1. The handler was introduced here https://github.com/ZEISS/precise-ui/pull/374. 
2. It gets passed down into the `PaginationLayout` here: https://github.com/ZEISS/precise-ui/blob/main/src/components/Pagination/index.tsx#L216
3. And added to a `div` here: https://github.com/ZEISS/precise-ui/blob/main/src/components/Pagination/PaginationLayout.part.tsx#L22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Precise UI Changelog
 
+## 2.1.15
+
+- Fix React error when onSizeChanged gets added to the Pagination component
+
 ## 2.1.14
 
 - Improve validation of `Form`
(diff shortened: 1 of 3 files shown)


klaidigorishti (approved):
LGTM - thanks for the fix!

PR #383, opened 2022-11-15
feat: make some component types compatible with react 18

feedm3:
### Prerequisites

### Contribution Type

### Description

We are using precise-ui in a react 18 project. Since react 18, `React.FC` doesn't have the `children` prop anymore. Therefore, we get a Typescript error as soon as we add a child to a component that uses `React.FC`.

In order to keep the change minimal, I did 2 things:
- Create a interface `ReactComponentDefaultProps` that can be used to extend property types for `React.Component`.
- Created a `ReactFC` interface, that can be used as a drop-in replacement for `React.FC`

I didn't replace all `React.FC` components with `ReactFC`, but only the ones we need right now to keep the change minimal.

### DIFF
--- a/src/components/Form/index.tsx
+++ b/src/components/Form/index.tsx
@@ -3,6 +3,7 @@ import styled from '../../utils/styled';
 import { Prompt } from '../Prompt';
 import { FormContext, FormContextType, FormValueNotifier, FormValueChange } from '../../contexts';
 import { StandardProps } from '../../common';
+import { ReactComponentDefaultProps } from '../../utils/react-18-compat';
 
 export interface FormSubmitEvent {
   /**
@@ -50,7 +51,7 @@ export interface FormValidationError {
   error: React.ReactChild;
 }
 
-export interface FormProps<FormValues> extends StandardProps {
+export interface FormProps<FormValues> extends StandardProps, ReactComponentDefaultProps {
   /**
    * Shows the given message if the user wants to navigate
    * with changes being made or renders custom component with message if provided.
@@ -351,6 +352,7 @@ export class Form<Values extends FormValuesData> extends React.Component<FormPro
       prompt,
       ...rest
     } = this.props;
+
     const { changed } = this.state;
     return (
       <StyledForm {...rest} onSubmit={this.submit}>
--- a/src/components/PaddedContainer/index.tsx
+++ b/src/components/PaddedContainer/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import styled, { css } from '../../utils/styled';
 import { distance } from '../../distance';
+import { ReactFC } from '../../utils/react-18-compat';
 
 export interface PaddedContainerProps {
   /**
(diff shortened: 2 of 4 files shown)


shirk33y (approved):

src/utils/react-18-compat.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+/**
+ * This interface needs to be added to a property type that is used on `React.Component`.
+ *
+ * With React 18, React.Component doesn't include children types anymore.
+ */
+export interface ReactComponentDefaultProps {
+  children?: React.ReactNode | undefined;
```
dremartins, line 9: isn't this redundant? `children?: React.ReactNode;` should be enough

src/components/Form/index.tsx
```diff
@@ -50,7 +51,7 @@ export interface FormValidationError {
   error: React.ReactChild;
 }
 
-export interface FormProps<FormValues> extends StandardProps {
+export interface FormProps<FormValues> extends StandardProps, ReactComponentDefaultProps {
```
dremartins, line 54: Why not add these to StandardProps?

src/components/PaddedContainer/index.tsx
```diff
@@ -40,6 +41,6 @@ const StyledContainer = styled('div')<PaddedContainerProps>(
   `,
 );
 
-export const PaddedContainer: React.FC<PaddedContainerProps> = props => {
+export const PaddedContainer: ReactFC<PaddedContainerProps> = props => {
```
dremartins, line 44: This looks too similar to the React.FC that it may cause confusion. Why not just extend PaddedContainerProps with the new props?

src/utils/react-18-compat.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/**
+ * This interface is a replacement for React.FC type to make it compatible with React 18 consumers.
+ *
+ * With React 18, React.FC doesn't include children types anymore.
+ */
+export interface ReactFC<G = {}> extends React.FC<G>, ReactComponentDefaultProps {}
```
dremartins, line 17: This name may cause confusion

src/utils/react-18-compat.ts
feedm3: This type is everywhere in the react official types and I wanted to stick with it:

screenshot-15-11-2022-at-15-47-07@2x

src/components/Form/index.tsx
feedm3: Will conflict with other interfaces that extend from them. 

For example, `children: void` is set in one of them.

src/utils/react-18-compat.ts
feedm3: Removed it, using `ReactComponentDefaultProps` only.

src/components/PaddedContainer/index.tsx
feedm3: Changed it.

dremartins (approved):

